PR #8, opened 2017-09-22
Add repetitionsPerFile and repetitionsPerFileRW options

glennklockwood:
These options allow more data to be read or written than the underlying volume allows by rewinding and re-writing/re-reading the same file over and over.


glennklockwood:
Whoops, wrong PR target.

PR #9, opened 2017-09-27
Changed the semantics of -R to compare the data with the expected dat…

JulianKunkel:
Previously -R did run another read phase and compared if the output of both reads is identical.
-R now checks if the data matches the expected signature (as set using -G <NUMBER>), so reads data once and then
directly compares the read data with the expected buffer.
This allows to first run IOR with a write only phase, then later with a read phase that checks if the data still is correct.
Since the read can be repeated multiple times, there is no need for the old -R semantics.

I removed quite some code that is now not longer needed.


johnbent:
This looks like a great patch but when I run with it, it's a bit slower than without.  I'm just on a local laptop and I'm not doing a ton of data so maybe it's just the extra memory operation that's causing the different.

PR #16, opened 2017-11-19
fix a bug when make IOR in MacOS will cause a compile error 

lwh1990:
The MacOS doesn't support lseek64 and open64. So when compiling IOR in MacOS will cause error "Undefined symbols for _lseek64 and _open64" in file contrib/cbif.c.

### DIFF
--- a/contrib/cbif.c
+++ b/contrib/cbif.c
@@ -36,6 +36,14 @@
 
 #define BYTE_BOUNDARY 8
 
+#ifndef   open64                /* necessary for TRU64 -- */
+#  define open64  open            /* unlikely, but may pose */
+#endif  /* not open64 */                        /* conflicting prototypes */
+
+#ifndef   lseek64               /* necessary for TRU64 -- */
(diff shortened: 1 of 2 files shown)


lwh1990:
fix another bug when compile with ncmpi enabled. The file "src/aiori-NCMPI.c" uses numTasksWorld as the process which is declared in "src/ior.h".   In "src/mdtest.c", NCMPI backend will be called if IOR configured with ncmpi enabled but numTasksWorld was not defined in "src/mdtest.c". So it will cause a compiler error like below:
mdtest-aiori-NCMPI.o: In function `NCMPI_Xfer':
/home/parallels/Documents/ior/src/aiori-NCMPI.c:272: undefined reference to `numTasksWorld'

glennklockwood:
Thanks for contributing these.  There's more safe ways to deal with the open64/lseek64 issues using autoconf and `_FILE_OFFSET_BITS`, but if someone runs into largefile issues on 32-bit, we can address this issue at that time.

PR #17, opened 2017-11-20
change MPI_BYTE to MPI_SIGNED_CHAR

lwh1990:
From Pnetcdf 1.7, the MPI datatype corresponding to NC_BYTE  is change from MPI_BYTE to MPI_SIGNED_CHAR. If not change, running IOR with NCMPI will cause a fatal error as below:
ERROR in aiori-NCMPI.c (line 287): cannot write to data set.
ERROR: NetCDF: Not a valid data type or _FillValue type mismatch.

### DIFF
--- a/src/aiori-NCMPI.c
+++ b/src/aiori-NCMPI.c
@@ -276,26 +276,26 @@ static IOR_offset_t NCMPI_Xfer(int access, void *fd, IOR_size_t * buffer,
         /* access the file */
         if (access == WRITE) {  /* WRITE */
                 if (param->collective) {
-                        NCMPI_CHECK(ncmpi_put_vara_all
+                        NCMPI_CHECK(ncmpi_put_vara_schar_all
                                     (*(int *)fd, var_id, offset, bufSize,
-                                     bufferPtr, length, MPI_SIGNED_CHAR),
+                                     bufferPtr),
                                     "cannot write to data set");
                 } else {
-                        NCMPI_CHECK(ncmpi_put_vara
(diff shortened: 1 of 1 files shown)


glennklockwood:
My intuition is that we should support both the earlier pre-1.7 and newer post-1.7 data types, but I don't keep up to date with the pnetcdf community and am unsure if there is a preferred way to address this within pnetcdf itself.  The safest bet is to detect the pnetcdf version via autoconf and set the appropriate data type at configure time.

I'll look into more general solutions and get back to this in the near future.

glennklockwood:
Maybe a dumb question, but is it possible to just use `NC_BYTE` directly within `aiori-NCMPI.c` so we don't have to do a bunch of preprocessor gymnastics every time this data type changes?

lwh1990:
In Pnetcdf, NC_BYTE is need by ncmpi_def_var() which is used to define the data variable. But in IO function such as ncmpi_put_vara() used in IOR will need corresponding MPI datatype as one of its parameter. Maybe there is another solution which doesn't need to provide MPI datatype. Pnetcdf does provide another kind of IO functions like ncmpi_put_var_schar() which do not need MPI datatype as parameter and do the same thing as ncmpi_put_vara(..MPI_SIGNED_CHAR ...). But I am not sure whether old release have support these funtion. I can try and test different version release. If it works, I will pull a request.

lwh1990:
I have added another commit which doesn't need MPI datatype parameter and tested. It works fine with Pnetcdf-1.4.0 (oldest release I can download in official website) and Pnecdf-1.8.1 (newset stable release).

PR #18, opened 2017-11-22
Hardens test for MPI

ikirker:
Previously, the configure script could fail to find an MPI
implementation but continue regardless. This alters the behaviour so
that the configure script fails if the check for MPI does.


glennklockwood:
Thanks, this is great.  Verified in the presence/absence of `MPI_init` with autoconf 2.67 and 2.69.

Should there ever come a day when we want IOR to build in the abscence of MPI, we can revisit this.

PR #31, opened 2018-01-25
Fix O_DIRECT option

ezickler:
The O_DIRECT option was not working as set_o_direct_flag() were moved to
utilities.c but there the  #define _GNU_SOURCE where missing. This lead
to the Warning "cannot use O_DIRECT".


glennklockwood:
Thanks.  Will keep #31 open to address the `__GNU` issue.

PR #33, opened 2018-02-14
Option to output results in a json file

ezickler:
This adds an option to output the results in a json file. The new
option is available through the 'jsonFileName' directive. For this to be
present IOR must be configured/compiled with '--with-json' flag. This
adds json-c as dependency and is therefore optional at compile time.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -205,10 +205,22 @@ Consider --with-aws4c=, CPPFLAGS, LDFLAGS, etc])
 ])
 
 
+# Enable json output
+PKG_PROG_PKG_CONFIG
 
-
-
-
+AC_ARG_WITH([json],
+  [AS_HELP_STRING([--with-json],
+    [build with json for output @<:@default=check@:>@])],
+  [],
+  [with_json=check])
+AS_CASE(["$with_json"],
+  [yes], [PKG_CHECK_MODULES([json], [json-c], [HAVE_JSON=1])],
+  [no], [],
+  [PKG_CHECK_MODULES([json], [json-c], [HAVE_JSON=1], [HAVE_JSON=0])])
+AM_CONDITIONAL([USE_JSON], [test "$with_json" != no -a "$HAVE_JSON" -eq 1])
+AM_COND_IF([USE_JSON],[
+        AC_DEFINE([USE_JSON], [], [Build with JSON ouput option])
+])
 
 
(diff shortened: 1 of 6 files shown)


glennklockwood:
#46 added a few iops-related counters back to the output of IOR.  It would be great if this json could use that iops calculation function and report that in the output as well.

That said, the acceptance of this PR should not block on adding iops output to the json.  We can update the json output after the PR is accepted.

ezickler:
Added the iops and fix an issue where I used int instead of long long in the json lib.

Regarding my question on guide lines in  #48, i would propose that we should discuss the chosen strings for the json objects. Actually I think they could be better. I would like to find a good rule for naming them. 

Some ideas:
* camel case like variables
* same names as internal variables  
  * this can be difficult as the are not always good in terms of self description 
  *  some do not have internal name like bw and iops

Some feedback on this would be welcome.

JulianKunkel:
I like to have JSON output, however, I would advise not to use a JSON library to reduce the dependencies for IOR and additional configure checks. 
JSON is trivial to create for our cases, such that we can just write it out using STDIO, hence preventing dependency checks and configure flags.
The library have benefits, when one tries to manipulate objects further, which is in this case unneeded.

Anyone against simplifying this?

JulianKunkel:
JSON output based on suggestion is available in the testing branch.
Example run:
./src/ior -i 3  -O summaryFormat=JSON  

Example output here:
[test.txt](https://github.com/hpc/ior/files/2173882/test.txt)
Combine with -O summaryFile to store the resulting JSON file.

Note: verbose output shall still be output to stdout, hence, combining JSON output with "-v" requires -O summaryFile to prevent garbage to be included in the JSON.

ezickler:
I'm with you on the dependency part, but on the other hand i don't like to always reimplement everything. It hurts the maintainability, cost time and can lead to errors . What do you think of static linking via submodules as a compromise?

JulianKunkel:
It was actually quite simple and was just like 50 lines to get it done,
which is IMHO ok to get rid of a dependency. Also allowed to reuse parts of
the code path to be shared between alternative implementations (default vs.
Json). Hence, would close the ticket.


Enno Zickler <notifications@github.com> schrieb am Mo., 9. Juli 2018, 23:51:

ezickler:
Didn‘t noticed it is that small. Nevermind my concernce than.

PR #34, opened 2018-02-21
Allow block sizes between 1 and 8 bytes

jyvet:
IOR_size_t was defined as long long which prevents using
block sizes below 8 bytes. This patch makes small changes
to be able to run IOR with blocks between 1 and 8 bytes.

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -362,8 +362,8 @@ CompareBuffers(void *expectedBuffer,
         size_t i, j, length, first, last;
         size_t errorCount = 0;
         int inError = 0;
-        unsigned long long *goodbuf = (unsigned long long *)expectedBuffer;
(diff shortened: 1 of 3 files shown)


glennklockwood:
I don't think this is a safe change to make.  There are a lot of places where `long long` is used instead of `IOR_size_t` and `void *` is used instead of `IOR_size_t *`.  Changing the typedef of something as fundamental as `IOR_size_t` (to be only 1 byte instead of 8!) without making sure that everything that implicitly assumes the 64-bit size of `IOR_size_t` is not something I feel comfortable accepting for an otherwise uncommon feature.

I'd love to revisit this at a time when IOR's data types are applied more consistently throughout the code, but right now I think the only thing that makes this patch not break for `transferSize` > 2^8 is a bunch of improperly typed stuff scattered throughout IOR.

Issue #35, opened 2018-02-23
IOR assumes a POSIX backend when removing files


shanedsnyder:
IOR calls `access()` to check a file's existence prior to calling into the backend to delete the file. This is unnecessary as the actual `unlink()` (or whatever the underlying call is) can often tell you this itself. At the very least, the `access()` call could be moved into backends that operate on POSIX file systems. Other backends will have their own mechanisms for verifying existence before removal, if needed.

JulianKunkel:
I believe this issue had been fixed with the refactoring.

PR #36, opened 2018-02-26
break the dependence on access in file remove path

shanedsnyder:
IOR now just calls the backend delete() function without doing
any sort of check for whether the file exists, leaving this as
a responsibility for the backend. each POSIX FS-based backend
was modified to account for this change:

- POSIX, HDF5, and NCMPI backends simply remove the file using
  unlink and catch any resulting ENOENT errors.
- MPI-IO is a little more tricky as it is not as straightforward
  to catch ENOENT errors when removing nonexistent files.


shanedsnyder:
This would resolve issue #35

shanedsnyder:
A few other comments about this merge:

- getting `access()` calls out of ior.c is appealing as it removes the assumption that all backends are POSIX-based
- `access()` does not tell you anything that you can not learn from calling `unlink()`
- `access()` is not a portable way for verifying file existence, due to weird prefixes sometimes tacked on to filenames for parsing by ROMIO (e.g., pvfs2:/path/to/testFile)

shanedsnyder:
Actually, I just made pretty sweeping changes to the pull request as I did not notice there was already an effort underway to abstract `access` calls (among others) inside of the aiori interface.

I basically backed out all of my original changes and committed the following:
- Renamed and exported all of the generic `aiori_` calls that assume a POSIX-based backend (e.g., `aiori_access()`, `aiori_mkdir()`, etc.). If modules do not set their function pointers to the generic variants of these calls, or implement their own variant, a warning is printed out that IOR will assume a POSIX-based backend going forward.
- IOR's `RemoveFile()` call now uses the `backend_>access()` to check for a file's existence rather than assuming every backend can support generic `access()` calls.
- HDF5 & NCMPI backends now call the MPIIO backend routine for removing a file, since their respective interfaces don't have a corresponding remove call. This is more appealing than just calling `unlink()` as the MPI-IO layer is is closer proximally to these interfaces.
- Logic was added to all generic POSIX-based `aiori_` calls to strip any ROMIO prefixes from filenames (e.g., ufs:/path/to/Testfile, remove the leading ufs prefix) before operating on them. This fixes a longstanding bug where IOR could not remove test files that were create with any ROMIO prefixes.

shanedsnyder:
Just noticed PR #39 is closely related to this. Maybe the thing to do is to accept the changes in that PR first, then I can update my changes in the MPI-IO module to use that `MPIIO_Access()` call? HDF5 and NCMPI should probably use that same `access()` call just as they use the same mechanism for checking file size and for deleting a file as MPIIO.

glennklockwood:
OK, #39 is merged.  Could you rebase against that and make sure it's workable with the rest of your changes?

shanedsnyder:
I don't know what exactly I did, but I botched the merge trying to merge in all of the new changes from master. I'm going to close this pull request and resubmit a new one with a sane commit history.

PR #37, opened 2018-02-28
Add Travis-CI config file

ezickler:
This adds a .travis.yaml file to use travis-ci for continuous
integration.
It currently compiles IOR on an ubuntu 14.04 machine with
only the POSIX, MPIIO and HDF5 backends enabled. For Lustre, GPFS,
NCMPI, HDFS and S3 libraries needs to be compiled as they are not
present in the ubuntu repositories. Lustre and NCMPI should be easy when
ubuntu 16.04 is available at travis-ci. S3 could be compiled by source
which shouldn't be hard but needs to be done if one is interested.
Hadoop needs a more complex installation process. GPFS seams not
possible as it is proprietary and needs a license as far as i know.

Repo Owner needs to enable CI on travis-ci.org. It's free for open
source git hub repos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,38 @@
+language: c
+# xenial 16.04 not yet supported
+# dist: xenial
+before_install:
+    # For MPI
+    - test -n $CC && unset CC
+install:
+    - sudo apt-get -qq update
+    # Install MPI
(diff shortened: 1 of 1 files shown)


glennklockwood:
Thanks for doing this.  We've really needed it with the volume of changes being committed and this is a huge help for basic sanity checking.

Issue #21, opened 2017-11-30
Questions while writing new feature.


ezickler:
Hi,
i'm adding a new feature to IOR for benchmarking compression enabled filesystems. As IOR only can generate data which is uncompressible or super good to compress (repeating the data every 128bit).
Therefor other data is needed. 
My approach is to give the user the possibility to fill the buffer with his own data. As this also should be a bigger chunk of data to avoid to be to easy to compress an IO buffer of `transferSize` is not sufficient. My approach so far is to have an IO buffer of `blockSize`. Which then gives me the possibility to have different transfers and not only repeating the same. 
Also i want to keep to the IOR behavior as much as possible to make us of all the options. That's why i'm sticking to the transfer, blocksize, segment pattern.  So far so good.

Now the comes the problem / question. As the offset for the accesses are precomputed in the `offsetArray` there is no chance in `WriteOrReadSingle()` to get the according transfer as say aren't the same for every access now. Code wise i need offsets for the iobuffer to set before `backend->xfer()` happen. My question now is where i should save these offsets.
I would like to compute and store them in an additional column in the `offsetArray`.
But i'm not sure is this would interfere with anything else. Are there any objections on this?

Additionally i'm not sure if i understand the behavior of `GetOffsetArrayRandom()` correctly. Especially the `filePerProc` variant. For one file its doing the same as GetOffsetArray and mixing up everything in the end but for `filePerProc` there is this `if ((random() % test->numTasks) == pretendRank)`. I think this means not every transfer for every process is done but just 1/(numProc) in the mean. Is this correct on desired behavior?

In general is this the right way to communicate or is there any kind of dev mailing list for IOR? I think i will have some more questions and ideas while working on IOR and i would like to get them upstream here as there seams to be some active development here. 

Kind regards,
Enno

glennklockwood:
> As the offset for the accesses are precomputed in the offsetArray there is no chance in WriteOrReadSingle() to get the according transfer as say aren't the same for every access now.

As you have discovered, IOR was not designed to read or write complex data, so you're going to have to make nontrivial changes to the way it performs its core I/O to make it do what you want.  I'm personally not enthusiastic about this since enabling IOR to perform arbitrarily complex reads and writes...

1. seems to only be useful when testing performance under compressibility (are there other use cases that would make it easier to expand the scope of IOR?)
2. would probably make it gnarlier to do read/write checking, and that code path is currently in flux (see #14)

But I would welcome opinions from other members of the IOR community on this.

[...]

PR #38, opened 2018-02-28
Unit test with googletest

ezickler:
This adds the googletest unit test framework. It is linked statically,
compiled and run only when 'make check' is called. Googletest library
lives in a new folder 'libs' and is added as a git submodule form the
github repo https://github.com/google/googletest
Test are created and compiled in a new folder 'tests' to keep them
separate for release path.
Initially a test for 'getOffsetArraySequential' is implemented.
This is just a starting point to get more confidence in the code when
changing or adding features as i would like to have when touching core 
parts of IOR.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libs/googletest"]
+	path = libs/googletest
+	url = git@github.com:google/googletest.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "libs/googletest"]
 	path = libs/googletest
-	url = git@github.com:google/googletest.git
+	url = https://github.com/google/googletest.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,9 @@ install:
     # GPFS
     # NOTE: Think GPFS need a license and is therefore not testable with travis.
 before_script: ./bootstrap
-script: ./configure --with-hdf5 && make
+script:
+    - ./configure --with-hdf5 && make
+    - ./configure  && make && make check
 
 
 # notifications:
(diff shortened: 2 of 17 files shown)


JulianKunkel:
I'm a bit indifferent on this one.
Some thoughts: good to add more tests; but using a C++ framework for testing C code adds the dependencies on a C++ compiler compared to C testing frameworks like: https://cmocka.org/
or simply writing some simple executables or simply not using a testing framework at all.

Having tests for low-level functionality with the partial mess of code that exists for historical reasons feels a bit cumbersome. But for some it does make sense to have unit tests.
In respect to system tests, the library version of IOR/MDTest allows a test to run IOR/MDTest, then test if the expected files actually exist / are filled with the expected data. Maybe that is also useful.

ezickler:
For clarification: I implemented this to verify my code for #23 as i didn't found any documentation on how getOffsetArray() and getOffsetRandomArray() should work. Discussion on this can be found here #21.

I asked some people in the working group for sugestion for c unit test frameworks and the lead me to googletest. I also found [this](https://stackoverflow.com/questions/65820/unit-testing-c-code) stack overflow post and [this](https://meekrosoft.wordpress.com/2009/11/09/unit-testing-c-code-with-the-googletest-framework/) blog entry, where they also quoted [this](https://pragprog.com/titles/jgade/test-driven-development-for-embedded-c) book. Thought i did not read it this and the active development of googletest  convinced me to use it. 

Never the less i can confirm the concerns of using a quite big c++ library for c testing. It wasn't pleasant to integrate it into the build processes and testing of a big monolithic c program (mostly one big file) like IOR wasn't either. I think a much more modularized code would be needed for good testing. Which i think would benefit the the code quality in general by the way.
I made this pull request as i thought better this than nothing. If some one has experience with other frameworks and thinks it fits better just drop this. @JulianKunkel  is this the case for cmocka ? 

I would be interested in automated (unit) tests via travis-CI.

JulianKunkel:
I'd close this for now. I would not add the dependency to a c++ tool.

PR #42, opened 2018-03-15
New documention with sphinx and Doxygen for readthedoc.io

ezickler:
These commits setup a sphinx and doxygen instance. The sphinx instance contains high level documentation for users and developer, which for now basically consists of the old user guide file split up in categories and a beginners tutorial. Doxygen is used to create source code documention which will more and more useful by adding function comments while coding. 
Sphinx can be configured to read the doxygen content via breath, this is not working at the moment, but could be done with some effort and would enable sphinx to reference code functions with comments and links. For now shpinx only provides a link to the doxygen instance.  
Both together are configured in such a way that that readthedocs.io will auto generate and upload them when setup for this repo.


glennklockwood:
This is a really great addition.  Thanks for contributing the documentation infrastructure.

PR #49, opened 2018-04-24
fix output to be allinge with column headers

ezickler:
This aligns the columns of the long summer probably. As I like to discuss in #48 i'm not sure if this should be merged directly, as it alters the output and may causes incompatibility with user scripts for post processing. If this i not a concern i think this really help to read the output  thought the table gets pretty wide now with iops and space for every column.


johnbent:
I like what you’ve been saying about not breaking output because many of us probably have parsers. Having said that, I suspect that white space changes won’t break most parsers. I’d avoid using tabs though. Safe to assume that most shells use fixed width fonts.

JulianKunkel:
I applied a slightly modified patch to the testing branch.

PR #50, opened 2018-04-25
fix IOR access dependency

shanedsnyder:
I'm recycling this from my earlier pull request (#36) that I botched, though the net result in each case is basically the same. Just wanted to clean up the git history which was a mess.

To summarize, this pull request should resolve issue #35 by breaking IOR's dependence on the `access()` call which is used in the IOR's file delete code path. This resolves a longstanding problem of not being able to remove IOR output "files" on non-POSIX backends. 

Summary of the changes:
- Expose the generic `aiori_` calls for `access()`, `mkdir()`, etc. so that backends can opt-in to them explicitly. POSIX, MPIIO, HDF5, and NCMPI backends all now explicitly set which of these generic functions they want to use.
- Expose MPIIO backend routines for `access()` and `delete()`. Now, the HDF5 and NCMPI backends use these MPIIO routines directly rather than implementing their own or using generic ones offered in `aiori`. Similar to the `file_get_size()` call, it's easier/cleaner for the HDF5/NCMPI backends to rely on MPIIO to check whether a file exists or to delete a file.
- IOR now no longer uses `access()` directly to check for a file's existence before deleting, and instead calls the correspoding `backend->access()` routine.

I have tested using both NCMPI and HDF5 backends and things appear to be working fine.

### DIFF
--- a/src/aiori-HDF5.c
+++ b/src/aiori-HDF5.c
@@ -92,6 +92,7 @@ static void HDF5_Delete(char *, IOR_param_t *);
 static void HDF5_SetVersion(IOR_param_t *);
 static void HDF5_Fsync(void *, IOR_param_t *);
 static IOR_offset_t HDF5_GetFileSize(IOR_param_t *, MPI_Comm, char *);
+static int HDF5_Access(const char *, int, IOR_param_t *);
 
 /************************** D E C L A R A T I O N S ***************************/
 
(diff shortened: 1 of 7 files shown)


JulianKunkel:
Good stuff! 
Merged into testing: 0052bff3e9d0881edde283d64ac8abb1e0a37be8

Issue #45, opened 2018-04-13
mmap support?


johnbent:
Would it be difficult to add an mmap backend?  That'd be pretty awesome.

glennklockwood:
It would not; in fact, I have the read-only part of one written:

https://github.com/glennklockwood/ior_gkl/blob/posix-mmap/src/aiori-MMAP.c

The code was written before the mdtest merge though, so there are probably bits and pieces that need to be cleaned up.  The create/write part of the aiori, as well as support for different mmap modes (private, shared, anon, etc) would also need to be added.

glennklockwood:
Thanks to @lidongyang for contributing this.  I created #53 to track the documentation for this new backend, but I didn't want to block the pull request on that.

PR #52, opened 2018-05-11
Add mmap as a new AIORI backend

lidongyang:
It shares the create/open/delete/set_version/get_file_size
functions with POSIX backend.
The mmap backend also supports fsync and fsyncPerWrite options,
and it will use msync() instead and fsync().

Signed-off-by: Li Dongyang <dongyangli@ddn.com>
Fixes: #45


glennklockwood:
This looks really fantastic.  Thank you so much for providing this functionality.

PR #54, opened 2018-05-16
use MPI_COMM_SELF when opening file to get file size in MPIIO

adammoody:
MPI requires all ranks to refer to the same file within the communicator passed to MPI_File_open.

This bug was exposed during testing with Open MPI and BeeGFS.  Without it, a run produces an error like the following:

srun -N2 -n4 ./install/bin/ior -a MPIIO -b 10MB -t 1MB -v -F -w -o /p/bscratcha/moody20/IOR_MANUAL

<snip>

Commencing write performance test: Wed May 16 13:22:31 2018
[catalyst105:75531] [1]mca_sharedfp_lockedfile_file_open: Error during file open
[catalyst106:37197] [2]mca_sharedfp_lockedfile_file_open: Error during file open
[catalyst106:37198] [3]mca_sharedfp_lockedfile_file_open: Error during file open

### DIFF
--- a/src/aiori-MPIIO.c
+++ b/src/aiori-MPIIO.c
@@ -464,8 +464,15 @@ IOR_offset_t MPIIO_GetFileSize(IOR_param_t * test, MPI_Comm testComm,
 {
         IOR_offset_t aggFileSizeFromStat, tmpMin, tmpMax, tmpSum;
         MPI_File fd;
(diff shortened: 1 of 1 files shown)


glennklockwood:
Yikes, good catch.  Thanks for the patch.

PR #56, opened 2018-05-16
open file in RDWR when using write + checkWrite

adammoody:


adammoody:
Seemed to be a duplicate "params->openFlags & IOR_RDWR" in the if statement that I also dropped.

glennklockwood:
Thanks for contributing these fixes!

adammoody:
You bet.  Glad to help!

PR #57, opened 2018-06-20
Add median MiB and OPs results

jyvet:
Context: Detailed summary displays the mean over all
iteratons for every operations. The mean may vary
a lot between multiple runs.

This patch adds the median in the detailed summary.

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -258,7 +258,7 @@ DisplayOutliers(int numTasks,
                 char *timeString, int access, int outlierThreshold)
 {
         char accessString[MAX_STR];
-        double sum, mean, sqrDiff, var, sd;
+        double sum, mean, median, sqrDiff, var, sd;
 
         /* for local timerVal, don't compensate for wall clock delta */
         timerVal += wall_clock_delta;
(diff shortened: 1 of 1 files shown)


JulianKunkel:
Thanks for submitting the suggestions.
I find it cumbersome to add more and more statistics, particularly when they involve more code and sorting. Instead I'd prefer we output the individual values and have users compute whatever they want (Median, Q90), who knows... Don't know what do the others think?

glennklockwood:
I tend to agree with @JulianKunkel on this.  As long as the results from each iteration are printed, users can calculate whatever percentile they'd like.  Ideally #33 (json output) would simplify the process for downstream analysis to do these sorts of calculations so that IOR itself doesn't have to present them.

jyvet:
Thanks for your comments. I like the idea of having a JSON output. Maybe, it would also help to simplify current IOR code to remove some stats and improve code clarity. If so, it would be neat to embed a script (python?) which parses that output, computes and displays basic statistics.

JulianKunkel:
@Jean-Yves, that would be great!
Please have a look at the testing branch, I believe the JSON output (a
minor revision happened yesterday) should fit your needs.
Python script would be nice!

How to use it is described here: https://github.com/hpc/ior/pull/33
(As said, the format of the file slightly changed, but gives an impression)

PR #71, opened 2018-08-09
Fix warnings reported at compilation time

jyvet:
This patch fixes most of the warnings caused
by unused variables and assignments from incompatible
type.


glennklockwood:
Thanks for submitting this clean-up work.  Changes look good.

PR #72, opened 2018-08-09
Fix behavior of delay between repetitions

jyvet:
Context: The help output states the -d parameter
introduces delay (in seconds) between repetitions.
But the delay is actually added before each operation,
and even before the very first iteration.

This patch makes the delay happen between repetitions.
It also removes the unecessary delay before the very
first iteration.

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -1242,6 +1242,10 @@ static void TestIoSys(IOR_test_t *test)
                 if (params->multiFile)
                         params->repCounter = rep;
 
+                /* Delay between repetitions */
+                if (rep > 0)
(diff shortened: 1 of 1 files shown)


jyvet:
Do you think delays between write and read tests (in the same iteration) are useful? If so I'll revisit the patch so that we can choose one (delay between each test, except before very first run) or the other behavior (delay between each iteration).

JulianKunkel:
Thanks. IMHO the delays between operations are useful. Sometimes caches are still flushed back after a write. Even though one uses fsync() or similar. Hence it may fix broken systems. I do not know if people use it for that purpose, though...

PR #77, opened 2018-08-17
Make Github properly detect the license

jyvet:
This adds a GPL LICENSE file.


glennklockwood:
I would prefer to keep only the terms and conditions of the GPL license as originally included in the `COPYRIGHT` file and drop the "Preamble" and everything after "END OF TERMS AND CONDITIONS" from `LICENSE`.  I don't think it was the intention of the original authors or the fundors of this work to pass judgment about "[taking] away your freedom" and would prefer all that language be dropped.

That said, happy to defer to the other maintainers if I am a minority in this.

jyvet:
Thanks for your feedbacks. I've done some tests, the license is still detected if we remove everything after "END OF TERMS AND CONDITIONS". However, it does not work anymore if the "Preamble" section is deleted. 

I can drop the patch if you prefer. It's not a big deal.

glennklockwood:
That's a shame.  It looks like GitHub uses a Ruby-based license detection library as mentioned here:

https://help.github.com/articles/licensing-a-repository/#detecting-a-license

I haven't had a chance to look over the GPL detector code and figure out exactly what it's looking for, but there may be a way to get this all to work.

Regardless of whether we can make it work or not, I appreciate you contributing this idea @jyvet.  It would be nice to have IOR show up as GPL-licensed project via the GitHub search API.

PR #83, opened 2018-09-18
Initialize only the required backends

jyvet:
Context: IOR initializes all available backends. If one
backend fails to initialize IOR cannot be used.

This patch makes IOR initialize only the backends
which will be used. The initialization is done after
that the parameters are checked so that the help message
can still be dispayed is something goes wrong.


JulianKunkel:
I like this a lot. Thanks

PR #84, opened 2018-09-18
Fix bug in computing max write and read bandwidths

jyvet:
This patch fixes a bug when computing the maximum
bandwidths in short summary output.


JulianKunkel:
Thanks, but besides variable renaming I cannot see any change, did you miss something to commit?

jyvet:
Line 628-629
```
        max_write = results[0].writeTime;
        max_read = results[0].readTime;
```
`max_write` and `max_read` are initialized with times instead of Bytes/s. Then for the first loop iteration it compares variables with different units (max_write is in seconds and bw is in Bytes/s) which
makes the MAX value wrong.

PR #101, opened 2018-10-08
compile correctly with -std=c99

### LINKED ISSUE #93: ior does not build without CFLAGS=-std=gnu99
A number of recent changes have essentially required that

1. the code be compiled with C99 support, as there are now pieces that violate c89 (e.g., `for (int i = 0; ...)`
2. there appear to be problems with certain gnuisms at link time (not sure if this is a `#ifdef _GNU_SOURCE` type of issue or what)

In particular, I am having lots of problems compiling ior's master using the Cray MPI environment (CLE6) and the Intel compiler, but I don't have the time to debug right now.  In the long term, we should decide if we are OK with requiring C99 (and if so, add that to autoconf).

My personal preference is to use strict c89 and add that to the regression tests' compiler flags so we catch non-compliant C.

glennklockwood:
This patch enables correct compilation on both MacOS and Linux using only POSIX.1-2008-compliant C with XSI extensions.  It also changes the travis build to explicitly use `-std=c99` to catch noncompliant patches going forward.

This does _not_ require a user compile with strict POSIX conformance though, and the general expectation is that most users will continue to build with various GNUisms enabled by default (and in fact, this is required to compile with `--with-lustre`).  The intent here is just to curb any creeping dependence on a specific libc or compiler.

resolves #93

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -19,6 +19,17 @@ AM_INIT_AUTOMAKE([check-news dist-bzip2 gnu no-define foreign subdir-objects])
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 AM_MAINTAINER_MODE
 
+# Check for system-specific stuff
+case "${host_os}" in
+  *linux*)
+    ;;
+  *darwin*)
+    CPPFLAGS="${CPPFLAGS} -D_DARWIN_C_SOURCE"
+    ;;
+  *)
+    ;;
+esac
+
 # Checks for programs
 
 # We can't do anything without a working MPI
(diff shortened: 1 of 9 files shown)


JulianKunkel:
Maybe the _XOPEN_SOURCE could be added to the CFLAGS instead of adding it to multiple files? Otherwise, great that you spend the time to get this done.

glennklockwood:
My preference is to keep this in source (somewhere) since it is a feature macro reflecting the features of the source.  Should someone mess with the build process later on, the declaration of POSIX shouldn't go away.

That said, we can stick this in a config somewhere and be sure to load that before any other headers across all our source files.  That may be easier to manage than sticking it in each source file that requires it.  Would this be better?

JulianKunkel:
Yes, I believe we can move this safely to config.h

PR #103, opened 2018-10-09
IME: add support of mkdir/rmdir into the IME backend

sydidelot:
This patch adds the support of `rmdir/mkdir` syscalls into the IME backend (feature supported starting from IME 1.3).
In addition, this patch fixes an issue in function PrependDir(), which bypasses the backend's implementation of `rmdir/mkdir`.


sydidelot:
Is there any change I can get some reviews for this patch?

sydidelot:
@JulianKunkel Thanks a lot for this quick response!

PR #107, opened 2018-10-17
mdtest: Use GetTimeStamp() instead of MPI_Wtime()

marcvef:
IOR uses the `GetTimeStamp()` wrapper (in src/utilities.h) to allow `gettimeofday()` to be used for timings instead of `MPI_Wtime()` to calculate throughput. This commits adds the same logic to mdtest timings. By default `MPI_Wtime()` is still used. If `_NO_MPI_TIMER` is added to, e.g., src/utilities.h `gettimeofday()` is used instead.

Having this possibility is important because `MPI_Wtime()` is known to report wrong timings on some Xeon architectures when used with OpenMPI-2.0.2 for instance ([see here for further explanation of this issue](https://stackoverflow.com/questions/42270974/mpi-wtime-timer-runs-about-2-times-faster-in-openmpi-2-0-2)), resulting in compromised mdtest throughput.

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -179,7 +179,7 @@ void offset_timers(double * t, int tcount) {
         fflush( out_logfile );
     }
 
-    toffset = MPI_Wtime() - t[tcount];
+    toffset = GetTimeStamp() - t[tcount];
     for (i = 0; i < tcount+1; i++) {
         t[i] += toffset;
     }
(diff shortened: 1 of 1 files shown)


dirtyharrycallahan:
OpenMPI 2.0 is retired and OpenMPI-2.0.2 was release in early 2017.

JulianKunkel:
@dirtyharrycallahan  , what conclusion do you derive from that statement?
For me, it looks good to have a wrapper that can take care of whatever issues with the clock...

glennklockwood:
This seems like a good idea to me.  Thanks for contributing it!

Issue #119, opened 2018-12-19
mdtest computes wrongly with stonewall


JulianKunkel:
The computation of mdtest when using stonewall was wrong. It used the number of items instead of the actual achieved stonewall number!

To verify if your mdtest version is *still* affected, run:
$ mdtest -C -W 1 -u -L -F  -n 100000000
This will run mdtest for 1 second to create files, you can check if the reported value is too high as it will output as well:
"Continue stonewall hit min: X"
Then you should see that mdtest reports for  File creation approximately X, too. Not 100000000.

Sorry about this. Have to track where the issue has been introduced.

JulianKunkel:
The bug is fixed with 4297ccd7f2130887417402e299acca7d2cb1ef9c

PR #121, opened 2018-12-19
cherry-pick solution to #119 into rc branch

glennklockwood:
@JulianKunkel could you just double check that this cherry pick looks right?  I'd rather have a second set of eyes since this is going straight into rc (and therefore the 3.2 release).


JulianKunkel:
It looks correct. Thanks

Am Mi., 19. Dez. 2018, 23:33 hat Glenn K. Lockwood <notifications@github.com>
geschrieben:

PR #122, opened 2018-12-20
MDTest New option: print rate AND time; improves debugging.

JulianKunkel:
It is great to have mdtest print either rate or time, but for debugging it is useful to print the time alongside.


JulianKunkel:
Integrated the patch into the other pull request. Done.

Issue #112, opened 2018-11-20
mdtest stat a file which is not created


lee-fan:
The problem is that mdtest stat a file which is not created previously.

the command i was using: mdtest -C -T -F -V=3 -d /mnt/mdtest -i 3 -I 10 -z 1 -b 2 -L -u
The following is the output from mdtest:
V-1: Entering PrintTimestamp...
-- started at 11/20/2018 14:23:50 --

mdtest-1.9.3 was launched with 1 total task(s) on 1 node(s)
Command line used: mdtest "-C" "-T" "-F" "-V=3" "-d" "/mnt/mdtest" "-i" "3" "-I" "10" "-z" "1" "-b" "2" "-L" "-u"
V-1: Entering parse_dirpath...
V-1: Entering valid_tests...
api                     : POSIX
barriers                : True
collective_creates      : False
create_only             : True
dirpath(s):
	/mnt/mdtest
dirs_only               : False
read_bytes              : 0
read_only               : False
first                   : 1
files_only              : True
iterations              : 3
items_per_dir           : 10
last                    : 0
leaf_only               : True
items                   : 0
nstride                 : 0
pre_delay               : 0
remove_only             : True
random_seed             : 0
stride                  : 1
shared_file             : False
time_unique_dir_overhead: False
stone_wall_timer_seconds: 0
stat_only               : True
unique_dir_per_task     : True
write_bytes             : 0
sync_file               : False
depth                   : 1
V-3: main (before display_freespace): testdirpath is "/mnt/mdtest"
V-1: Entering display_freespace...
V-3: testdirpath is "/mnt/mdtest"
V-3: Before show_file_system_size, dirpath is "/mnt"
V-1: Entering show_file_system_size...
Path: /mnt
FS: 517.6 GiB   Used FS: 39.4%   Inodes: 32.9 Mi   Used Inodes: 0.5%
V-3: After show_file_system_size, dirpath is "/mnt"
V-3: main (after display_freespace): testdirpath is "/mnt/mdtest"

1 tasks, 30 files

   Operation               Duration              Rate
   ---------               --------              ----
V-1: main: * iteration 1 *
V-2: main (for j loop): making testdir, "/mnt/mdtest/#test-dir.0-0"
V-3: main (create hierarchical directory loop-!collective_creates): Calling create_remove_directory_tree with "/mnt/mdtest/#test-dir.0-0"
V-1: Entering create_remove_directory_tree, currDepth = 0...
[...]

PR #123, opened 2018-12-20
Fix 112 + new feature to compare patterns.

JulianKunkel:
This is a bugfix for #112.
It also introduces a new testing environment for mdtest that allows to compare expected patterns with observed patterns.
I think this is the way to move forward to change mdtest and ensure it works as expected.


glennklockwood:
Thanks for taking the time to add all these tests.

PR #133, opened 2019-02-11
Enable IME native backend with mdtest

jyvet:


src/mdtest.c
```diff
@@ -1573,7 +1573,9 @@ void valid_tests() {
         FAIL("-c not compatible with -B");
     }
 
-    if (strcasecmp(backend_name, "POSIX") != 0 && strcasecmp(backend_name, "DUMMY") != 0) {
+    if ((strcasecmp(backend_name, "POSIX") != 0) &&
+        (strcasecmp(backend_name, "IME")   != 0) &&
+        (strcasecmp(backend_name, "DUMMY") != 0)) {
       FAIL("-a only supported interface is POSIX (and DUMMY) right now!");
```
glennklockwood: Can you update the error message to include IME?

glennklockwood:
Thanks for contributing this!  The changes seem very well encapsulated.  I don't know if we have a way to test this independently, but it looks like it won't really affect non-IME functionality at all.

The error message should perhaps be updated, or maybe only updated if compiled with IME enabled.  Happy to hear opinions from others on this.

src/mdtest.c
jyvet: Sure I'll resubmit a version. Thanks for the review!

jyvet:
Yes we have a way to test this independently. We made a mock interface which fakes a IME client and does POSIX calls. I'll do another PR to incude this in the Travis tests.
Thanks

jyvet:
Hi, could you tell me if you see other stuff I could improve for those 2 patches?
Thanks!

src/aiori.h
```diff
@@ -83,8 +84,14 @@ typedef struct ior_aiori {
         void (*initialize)(); /* called once per program before MPI is started */
         void (*finalize)(); /* called once per program after MPI is shutdown */
         option_help * (*get_options)();
+        bool enable_mdtest;
 } ior_aiori_t;
 
+enum bench_type {
```
JulianKunkel: I find this solution a bit crude. AIORI at some point might become a library and useful to other benchmarks. But well, Ok for now.

PR #137, opened 2019-03-14
Fix segfault with option memoryPerNode

sydidelot:
optarg contains a NULL address, which causes a segfault in
NodeMemoryStringToBytes().


JulianKunkel:
Well spotted.

PR #142, opened 2019-04-10
Adding support for NetCDF-4

kmpaul:
I realize that NetCDF-4 is built on HDF5 (or PnetCDF, if enabled), and both HDF5 and PnetCDF are supported by IOR.  However, I routinely encounter issues related to the extra overhead that NetCDF add on top of HDF5, and for that reason I would like the ability to test NetCDF-4 along side HDF5, in the same framework.

I have noticed that testing seems a little light.  The only tests that appear to be run at the `testing/basic-tests.sh` script, which actually only tests the `POSIX` and `MMAP` APIs.  Are there plans to extend these tests?  I'm wondering how best to add tests for the new code in this PR (other than just testing if IOR can build with the `--with-nc4` configuration flag).

### DIFF
--- a/src/aiori-NC4.c
+++ b/src/aiori-NC4.c
@@ -1,0 +1,358 @@
+/* -*- mode: c; c-basic-offset: 8; indent-tabs-mode: nil; -*-
+ * vim:expandtab:shiftwidth=8:tabstop=8:
+ */
+/******************************************************************************\
+*                                                                              *
+*        Copyright (c) 2003, The Regents of the University of California       *
+*      See the file COPYRIGHT for a complete copyright notice and license.     *
+*                                                                              *
+********************************************************************************
+*
+* Implement abstract I/O interface for NetCDF-4 with parallel support (NC4).
+*
+\******************************************************************************/
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <string.h>
+#include <netcdf.h>
+#include <netcdf_par.h>
+
+#include "ior.h"
+#include "iordef.h"
+#include "aiori.h"
+#include "utilities.h"
+
+#define NUM_DIMS 3              /* number of dimensions to data set */
+
+/*
+ * NC4_CHECK will display a custom error message and then exit the program
+ */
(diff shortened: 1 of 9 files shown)


src/aiori-NC4.c
```diff
@@ -0,0 +1,355 @@
...
+    return (fd);
+}
+
+/*
+ * Write or read access to file using the NC4 interface.
+ */
+static IOR_offset_t NC4_Xfer(int access, void *fd, IOR_size_t *buffer,
+                             IOR_offset_t length, IOR_param_t *param) {
+    signed char *bufferPtr = (signed char *) buffer;
+    static int firstReadCheck = FALSE, startDataSet;
```
JulianKunkel: The use of a static variable here appears problematic. As IOR could repeat an experiment multiple times, this shouldn't be a static variable.

kmpaul: Thanks, @JulianKunkel!  I'll take a closer look at the logic here.  To be honest, I cribbed this code from the NCMPI aiori, which also uses a `static` declaration.  See:

https://github.com/hpc/ior/blob/d49d345007530a91e9f36dfdab87bb46c006a674/src/aiori-NCMPI.c#L176-L180

I think that Wei-keng Liao had intended to use this static variable to only enable certain code for the first call of `*_Xfer` on a `READCHECK`.  So, I think that logic requires it to be static.

However, I'm happy to rethink the logic here and implement a different solution, if you'd prefer.

JulianKunkel:
cool. I think a problem is the detection if that variable is the first time
used.
The check on Line 173:
if ((int) (param->offset - segmentPosition) == 0) { Might be problematic
for random accessed files. I'm not sure if that ever was used / tested
properly with NCMPI either :-) Did you check that it works for multiple
processes / iterations with random pattern?

kmpaul:
I have not tested random access!  Only sequential.  I'll give that a try.

Thanks for the tip, @JulianKunkel!

kmpaul:
@JulianKunkel Sorry for the delay.  Lots of stuff on my plate, and it took me some time to get back to this.

It appears that, like the HDF5 API, the NCMPI API was implemented to throw an error if random offsets are specified.  Hence, since I cribbed the NC4 API from the NCMPI API, the same is true.  I didn't think about that when you asked your question about random access last week.

PR #157, opened 2019-05-23
aiori-IME: Fix compilation by removing o_direct check

jyvet:
useO_DIRECT was removed from IOR_param_t structure preventing
IME backend from compiling. This patch removes the check on
O_DIRECT flag.

### DIFF
--- a/src/aiori-IME.c
+++ b/src/aiori-IME.c
@@ -130,9 +130,6 @@ static void *IME_Open(char *testFileName, IOR_param_t *param)
         if (fd == NULL)
                 ERR("Unable to malloc file descriptor");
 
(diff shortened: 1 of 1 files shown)


JulianKunkel:
Hi, can you have a look at:
1b3182904d9840b9542ab6ed9dbc6d9733078562
That preserves odirect and allows IME to have more options (if you offer them ^-^)

jyvet:
Hi, thanks for your patch! It works fine with IME. Indeed, it gives us the opportunity to use options with IME backend! For now with IME, IOs are always O_DIRECT. But I think we should use your patch since it may change in the future.

JulianKunkel:
Fixed with  #160

PR #158, opened 2019-05-23
add mknod support

KnightKu:
Add new option '-k' to support file creation test with mknod,
which is widely used in lustre.

Signed-off-by: Gu Zheng <gzheng@ddn.com>

### DIFF
--- a/src/aiori-POSIX.c
+++ b/src/aiori-POSIX.c
@@ -107,6 +107,7 @@ ior_aiori_t posix_aiori = {
         .name = "POSIX",
         .name_legacy = NULL,
         .create = POSIX_Create,
+        .mknod = POSIX_Mknod,
         .open = POSIX_Open,
         .xfer = POSIX_Xfer,
         .close = POSIX_Close,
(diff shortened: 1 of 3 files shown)


src/aiori-POSIX.c
```diff
@@ -427,6 +428,24 @@ void *POSIX_Create(char *testFileName, IOR_param_t * param)
...
 }
 
+/*
+ * Creat a file through mknod interface.
+ */
+void *POSIX_Mknod(char *testFileName)
+{
+	int *fd;
+
+	fd = (int *)malloc(sizeof(int));
```
JulianKunkel: This is not optimal, as with Mknod you do not call POSIX_close() which calls  free(fd). I will fix this, though in a subsequent patch.

JulianKunkel:
Now that I did some further testing, as you create regular files at the moment, these could support "-w" and File_close(). Or do you aim to use it to create special inodes and such?

KnightKu:
Hi @JulianKunkel , thanks for your quick review and improvement.
We only use it to create regular files at the moment.

PR #162, opened 2019-07-27
Fixed shift.  Also cleaned up output messages

johnbent:
If you look at the differences, they appear like a lot but the overwhelming majority are merely cleaning up output messages in mdtest.  It used to have at least 3 lines for every output message because it wrapped each one up in a if branch testing for verbosity.  I added a helper function so now it is just one line of code for each output message.

There there is also a new function added in utilities which prints the node map (a 0 or 1 for every rank where 0 means on a different node from rank 0 and 1 means same node) which also returns whether the node/task mapping is round-robin or contiguous.

Then a small change in mdtest to shift by ppn if contiguous or 1 otherwise.  It used to only shift by 1 so it used to only work when the mapping was round robin.  Now it works for both.

Then a small change in IOR to do the same.  It used to only shift by ppn so it used to only work when the mapping was contiguous.  Now it works for both.

I tested the code at small scale and Ken Carlilie (thanks Ken!) tested at large scale.

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -939,7 +939,7 @@ static void InitTests(IOR_test_t *tests, MPI_Comm com)
                 params->testComm = com;
                 params->nodes = params->numTasks / tasksPerNode;
                 params->tasksPerNode = tasksPerNode;
-                params->tasksBlockMapping = QueryNodeMapping(com);
+                params->tasksBlockMapping = QueryNodeMapping(com,false);
                 if (params->numTasks == 0) {
                   params->numTasks = size;
                 }
@@ -1223,7 +1223,7 @@ static void TestIoSys(IOR_test_t *test)
         }
         if (rank == 0 && params->reorderTasks == TRUE && verbose >= VERBOSE_1) {
                 fprintf(out_logfile,
-                        "Using reorderTasks '-C' (expecting block, not cyclic, task assignment)\n");
+                        "Using reorderTasks '-C' (useful to avoid read cache in client)\n");
                 fflush(out_logfile);
         }
         params->tasksPerNode = CountTasksPerNode(testComm);
--- a/src/ior.c
+++ b/src/ior.c
@@ -939,6 +939,7 @@ static void InitTests(IOR_test_t *tests, MPI_Comm com)
                 params->testComm = com;
                 params->nodes = params->numTasks / tasksPerNode;
                 params->tasksPerNode = tasksPerNode;
+                params->packedTasks = QueryNodeMapping(com);
                 if (params->numTasks == 0) {
                   params->numTasks = size;
                 }
@@ -1360,7 +1361,13 @@ static void TestIoSys(IOR_test_t *test)
                         }
                         if (params->reorderTasks) {
                                 /* move two nodes away from writing node */
-                                rankOffset = (2 * params->tasksPerNode) % params->numTasks;
+                                int shift = 1;
+                                if (params->packedTasks) {
+                                    shift = params->tasksPerNode;
+                                } else {
+                                    shift = 1;
+                                }
+                                rankOffset = (2 * shift) % params->numTasks;
                         }
 
                         // update the check buffer
@@ -1395,9 +1402,14 @@ static void TestIoSys(IOR_test_t *test)
                         /* Get rankOffset [file offset] for this process to read, based on -C,-Z,-Q,-X options */
                         /* Constant process offset reading */
                         if (params->reorderTasks) {
-                                /* move taskPerNodeOffset nodes[1==default] away from writing node */
-                                rankOffset = (params->taskPerNodeOffset *
-                                          params->tasksPerNode) % params->numTasks;
+                                /* move one node away from writing node */ 
+                                int shift = 1;
+                                if (params->packedTasks) {
+                                    shift=params->tasksPerNode;
+                                } else {
+                                    shift=1;
+                                }
+                                rankOffset = (params->taskPerNodeOffset * shift) % params->numTasks;
                         }
                         /* random process offset reading */
                         if (params->reorderTasksRandom) {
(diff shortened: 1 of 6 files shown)


johnbent:
By the way, looks like travis CI is failing but it has nothing to do with my commit:

aiori-HDF5.c:25:18: fatal error: hdf5.h: No such file or directory

For some reason, it is passing --with-hdf5 to ./configure.  Not sure how to fix that but it's independent of my pull request.

src/ior.h
```diff
@@ -100,6 +100,7 @@ typedef struct
     int numTasks;                    /* number of tasks for test */
     int nodes;                       /* number of nodes for test */
     int tasksPerNode;                /* number of tasks per node */
+    int packedTasks;                 /* are the tasks round-robin across nodes or are they packed tightly*/
```
JulianKunkel: maybe instead of "packed" the term mpiBlockMapping would be more appropriate?

src/ior.c
```diff
@@ -1395,9 +1402,14 @@ static void TestIoSys(IOR_test_t *test)
...
                         if (params->reorderTasks) {
-                                /* move taskPerNodeOffset nodes[1==default] away from writing node */
-                                rankOffset = (params->taskPerNodeOffset *
-                                          params->tasksPerNode) % params->numTasks;
+                                /* move one node away from writing node */ 
+                                int shift = 1;
+                                if (params->packedTasks) {
+                                    shift=params->tasksPerNode;
+                                } else {
+                                    shift=1;
```
JulianKunkel: can remove the else branch

src/ior.h
johnbent: Nice suggestion.  Changed.

src/ior.c
johnbent: Nice suggestion.  Removed.

glennklockwood:
I’ll take a look at this before end of Wednesday so we can get this merged in.

src/mdtest.c
```diff
@@ -2364,26 +2057,31 @@ mdtest_results_t * mdtest_run(int argc, char **argv, MPI_Comm world_com, FILE *
...
+    VERBOSE(3,-1,"main (before display_freespace): testdirpath is '%s'", testdirpath );
 
     if (rank == 0) display_freespace(testdirpath);
+    int tasksBlockMapping = QueryNodeMapping(testComm);
 
-    if (verbose >= 3 && rank == 0) {
-        fprintf(out_logfile,  "V-3: main (after display_freespace): testdirpath is \"%s\"\n", testdirpath );
-        fflush( out_logfile );
+    /* set the shift to mimic IOR and shift by procs per node */
+    if (nstride > 0) {
```
glennklockwood: Ideally the semantics surrounding the mdtest `-N` argument and the IOR `-C` argument would be similar, but they currently are not.  It makes me a little uncomfortable to make the behavior of `-N 1` depend on the way the MPI wireup maps ranks to processes, but I don't know a reasonable way to fix this.

That said, the help line in line 1865 should be updated to clarify that `-N 1` is not always a 1-task shift, but a 1-node shift when the resource manager does block allocation.

src/mdtest.c
```diff
@@ -1312,77 +1191,10 @@ void file_test(const int iteration, const int ntasks, const char *path, rank_pro
...
-               t[2] - t[1], summary_table[iteration].rate[5]);
-        fprintf(out_logfile, "V-1:   File read         : %14.3f sec, %14.3f ops/sec\n",
-               t[3] - t[2], summary_table[iteration].rate[6]);
-        fprintf(out_logfile, "V-1:   File removal      : %14.3f sec, %14.3f ops/sec\n",
-               t[4] - t[3], summary_table[iteration].rate[7]);
-        fflush(out_logfile);
-    }
-}
-
-void print_help (void) {
```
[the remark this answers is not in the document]
glennklockwood: nevermind, I see where this is now handled below.

glennklockwood:
I have no idea why Travis is failing on hpc/ior.  I ran John's fork through my own Travis-enabled clone and it passed:

https://travis-ci.org/glennklockwood/ior/builds/565774378

The error from Travis makes no sense, so either someone with a better understanding of Travis (and how to clean its state?) will have to take a look, or I'll need a few hours and a clearer mind to try to reproduce and fix whatever is ailing our CI.

At any rate, John's PR doesn't introduce any regressions.  Would be nice to have a test to validate the new `mdtest -N` behavior so it also doesn't regress, but beggars can't be choosers.

src/ior.c
```diff
@@ -939,6 +939,7 @@ static void InitTests(IOR_test_t *tests, MPI_Comm com)
                 params->testComm = com;
                 params->nodes = params->numTasks / tasksPerNode;
                 params->tasksPerNode = tasksPerNode;
+                params->tasksBlockMapping = QueryNodeMapping(com);
```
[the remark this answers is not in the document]
johnbent: OK.  I added a parameter to it to control whether it prints or not so that mdtest can continue to print it but IOR will not.

src/ior.c
```diff
@@ -1360,7 +1361,11 @@ static void TestIoSys(IOR_test_t *test)
                         }
                         if (params->reorderTasks) {
                                 /* move two nodes away from writing node */
-                                rankOffset = (2 * params->tasksPerNode) % params->numTasks;
+                                int shift = 1; /* assume a by-node (round-robin) mapping of tasks to nodes */
+                                if (params->tasksBlockMapping) {
+                                    shift = params->tasksPerNode; /* switch to by-slot (contiguous block) mapping */
```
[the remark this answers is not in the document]
johnbent: You are understanding this correctly!  Good catch.  Changed.

src/mdtest.c
johnbent: I think that the semantics are almost the same now.  -N 1 in mdtest is identical to -C in IOR.  I agree it would be better if they were identical.  It would be great if it was the same flag but -N is already used for something else in IOR and -C is already used for something else in mdtest.  We could remove the argument to -N in mdtest; that would actually make a lot of sense since I think the only reason it was there was to allow users to explicitly set it to PPN for block allocation.  Want me to do that?

johnbent: Changed the help line.  Thanks!

PR #167, opened 2019-08-01
Feature verify mdtest

JulianKunkel:
That branch allows to use the -X parameter for MDTest which verifies that the data read (with -e) is the expected pattern.
MDTest does already check that the bytes read are the right number of bytes, but not 
(Note: build upon the feature-mdtest-timer branch)

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -700,6 +700,7 @@ void mdtest_read(int random, int dirs, const long dir_iter, char *path) {
 
         /* read file */
         if (read_bytes > 0) {
+            read_buffer[0] = 42;
             if (read_bytes != (size_t) backend->xfer (READ, aiori_fh, (IOR_size_t *) read_buffer, read_bytes, &param)) {
                 FAIL("unable to read file %s", item);
             }
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -131,6 +131,7 @@ static uint64_t items_per_dir;
 static uint64_t num_dirs_in_tree_calc; /* this is a workaround until the overal code is refactored */
 static int directory_loops;
 static int print_time;
+static int print_rate_and_time;
 static int random_seed;
 static int shared_file;
 static int files_only;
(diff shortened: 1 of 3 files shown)


src/mdtest.c
```diff
@@ -682,9 +700,16 @@ void mdtest_read(int random, int dirs, const long dir_iter, char *path) {
 
         /* read file */
         if (read_bytes > 0) {
+            read_buffer[0] = 42;
```
johnbent: What is 42?  Can we make it an enum typedef or something?

src/mdtest.c
```diff
@@ -1210,129 +1238,141 @@ void summarize_results(int iterations) {
...
-
-                if (min > curr) {
-                    min = curr;
-                }
-                if (max < curr) {
-                    max =  curr;
+            for (k=0; k < size; k++) {
+                for (j = 0; j < iterations; j++) {
+                    curr = all[(k*tableSize*iterations)
+                               + (j*tableSize) + i];
```
johnbent: I wonder if these two lines should be combined.  Doing so pushes beyond 80 chars on the line?  Do we have a style guide for IOR coding conventions?

src/mdtest.c
```diff
@@ -1210,129 +1238,141 @@ void summarize_results(int iterations) {
...
+                    if (min > curr) {
+                        min = curr;
+                    }
```
johnbent: Similarly, I wonder if these if statements with just a single op should be one-liners?  Just makes code more condense so easier to scan it quickly.

src/mdtest.c
JulianKunkel: It is just to overwrite the read buffer with something that is not in the pattern. It just cannot be 0.

src/mdtest.c
JulianKunkel: Surely, that was all the code that existed, I did change the condition to simplify the code, next time, improve that pieces :-)

PR #169, opened 2019-08-02
Fix: Heap Buffer Overflow

ax3l:
Fix a memory violation when run in serial.

Found via gcc's `-fsanitize=address` sanitizer.


ax3l:
Oh yes, fully agree. In fact, I was a bit confused why it reduces to index `1` anyway.

ax3l:
Not sure if I complicate the source further with early returns since also some logs are written here. Probably it's clearer already with the last push?

JulianKunkel:
I think this is alright, given the simplicity... Good catch.

PR #170, opened 2019-08-03
Fix Some Memory Leaks

ax3l:
Fixing some memory leaks :)

Found via gcc's `-fsanitize=address` sanitizer. (There are more, but that's what I catched so far ;) )

### DIFF
--- a/src/aiori-POSIX.c
+++ b/src/aiori-POSIX.c
@@ -89,6 +89,7 @@ option_help * POSIX_options(void ** init_backend_options, void * init_values){
   }
 
   *init_backend_options = o;
+  free(o);
 
   option_help h [] = {
     {0, "posix.odirect", "Direct I/O Mode", OPTION_FLAG, 'd', & o->direct_io},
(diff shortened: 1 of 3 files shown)


ax3l:
Ah dang, I added a mem violation in that one. Will add another PR.

ax3l:
There is so much more in option parsing, but I can not get a real hold of it. E.g. 

```
diff --git a/src/parse_options.c b/src/parse_options.c
index af30c36..362443d 100755
--- a/src/parse_options.c
+++ b/src/parse_options.c
@@ -550,6 +550,16 @@ IOR_test_t *ParseCommandLine(int argc, char **argv)
     option_parse(argc, argv, global_options);
     updateParsedOptions(& initialTestParams, global_options);
 
+    for (int i=1; i< global_options->module_count; i++) {
+      if(global_options->modules[i].options != NULL){
+          free(global_options->modules[i].options);
+      }
+    }
+    free(options);
+
+    free(global_options->modules);
+    free(global_options);
+
     if (initialTestParams.testscripts){
       tests = ReadConfigScript(initialTestParams.testscripts);
     }else{
```

PR #173, opened 2019-08-04
Fix Last Free

ax3l:
Fix bug introduced in #170.

Cannot get a good hold on the memory lifetime of the parsed options and the memory leaks they introduce. So I just leave them as is.


ax3l:
ping @JulianKunkel @glennklockwood

PR #177, opened 2019-08-30
Workaround for #24.

jschwartz-cray:
This issue is still being discussed but in the meantime I propose we remove this code until a more correct solution for S3 can be found since it breaks others.


jschwartz-cray:
Closing this PR in light of c83edfe on #177.

PR #182, opened 2019-08-30
Fix #181.

### LINKED ISSUE #181: tasksPerNode can vary between nodes
While tasksPerNode is commonly the same across every node, it's not guaranteed.  Presently IOR uses this value to reorder tasks and on systems where this property isn't true they aren't reordered properly.

For example, if you have 64 tasks and 4 nodes which can run 20 tasks each you get 3 groups of 20 and 1 group of 4.  On the groups of 20, tasksPerNode is 20, and on the group of 4, tasksPerNode is 4, resulting in the group of 4 nodes reordering their tasks differently.

The result is some nodes which read multiple files, and others which read none.

I've got some proposed changes for this issue which I'll be submitting as a PR momentarily.

jschwartz-cray:
On systems where numTasks is not evenly divisible by 'tasksPerNode' we were
seeing some nodes reading multiple files while others read none after
reordering.

Commonly all nodes have the same number of tasks but there is nothing
requiring that to be the case.  Imagine having 64 tasks running against 4
nodes which can run 20 tasks each.  Here you get three groups of 20 and one
group of 4.  On this sytem nodes running in the group of 4 were previously
getting tasksPerNode of 4 which meant they reordered tasks differently than
the nodes which got tasksPerNode of 20.

The key to fixing this is ensuring that every node reorders tasks the same
way, which means ensuring they all use the same input values.  Obviously on
systems where the number of tasks per node is inconsistent the reordering will
also be inconsistent (some tasks may end up on the same node, or not as far
separated as desired, etc.) but at least this way you'll always end up with a
1:1 reordering.

- Renamed nodes/nodeCount to numNodes
- Renamed tasksPerNode to numTasksOnNode0
- Ensured that numTasksOnNode0 will always have the same value regardless of
  which node you're on
- Removed inconsistently used globals numTasksWorld and tasksPerNode and
  replaced with per-test params equivalents
- Added utility functions for setting these values:
  - numNodes -> GetNumNodes
  - numTasks -> GetNumTasks
  - numTasksOnNode0 -> GetNumNodesOnTask0
- Improved MPI_VERSION < 3 logic for GetNumNodes so it works when numTasks is
  not evenly divisible by numTasksOnNode0
- Left 'nodes' and 'tasksPerNode' in output alone to not break compatibility
- Allowed command-line params to override numTasks, numNodes, and
  numTasksOnNode0 but default to using the MPI-calculated values

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -120,7 +120,7 @@ Version 2.10.1
 - Corrected IOR_GetFileSize() function to point to HDF5 and NCMPI versions of
   IOR_GetFileSize() calls
 - Changed the netcdf dataset from 1D array to 4D array, where the 4 dimensions
-  are: [segmentCount][numTasksWorld][numTransfers][transferSize]
+  are: [segmentCount][numTasks][numTransfers][transferSize]
   This patch from Wei-keng Liao allows for file sizes > 4GB (provided no
   single dimension is > 4GB).
 - Finalized random-capability release
--- a/src/aiori-NCMPI.c
+++ b/src/aiori-NCMPI.c
@@ -216,7 +216,7 @@ static IOR_offset_t NCMPI_Xfer(int access, void *fd, IOR_size_t * buffer,
                             param->blockSize / param->transferSize;
 
                         /* reshape 1D array to 3D array:
-                           [segmentCount*numTasksWorld][numTransfers][transferSize]
+                           [segmentCount*numTasks][numTransfers][transferSize]
                            Requirement: none of these dimensions should be > 4G,
                          */
                         NCMPI_CHECK(ncmpi_def_dim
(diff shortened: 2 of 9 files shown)


jschwartz-cray:
I did my best to follow the surrounding style/whitespace but there were places where it wasn't consistent so I used my best judgement.  I'm happy to fix anything that stands out, please just let me know.

A few notes and things I would like feedback / a second set of eyes on:
- The `"WARNING: More tasks requested (%d) than available (%d),"` message which is now in `InitTests` seems unnecessary.  I kept it in my changes as it was in the code previously, but overriding -N (numTasks) on the command-line only affects how things get reordered so setting a larger value shouldn't be a problem if someone wants to test with that.  Though it also doesn't seem to add much value either.
- The `params->tasksPerNode = CountTasksPerNode(testComm);` that was in `TestIoSys` did not seem to serve any purpose other than 1) doing it for every test (slow) and 2) preventing the value from being overridden on the command-line.  `InitTests` will always be called before this point and that method is ensuring this is set correctly.

jschwartz-cray:
Any thoughts on this PR @JulianKunkel?  Would you like to see anything changed?  I'd like to see this bug get fixed before the next release.

JulianKunkel:
I do believe that one may need some testing, i.e., different MPI versions and such, but yeah good for inclusion to the main (dev) branch right now.

PR #184, opened 2019-09-01
Feature aiori

JulianKunkel:
Changes to aiori supporting sync #175, fixing #24, #177

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -1677,15 +1677,12 @@ static void ValidateTests(IOR_param_t * test)
         if (test->useExistingTestFile && test->lustre_set_striping)
                 ERR("Lustre stripe options are incompatible with useExistingTestFile");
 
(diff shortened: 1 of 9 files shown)


src/ior.c
```diff
@@ -1677,15 +1677,12 @@ static void ValidateTests(IOR_param_t * test)
...
-        IOR_offset_t  b    = test->blockSize;
-
-        if (Nto1 && (s != 1) && (b != t)) {
-                ERR("N:1 (strided) requires xfer-size == block-size");
+        /* allow the backend to validate the options */
+        if(test->backend->check_params){
+          int check = test->backend->check_params(test);
+          if (check == 0){
+            ERR("The backend returned that the test parameters are invalid.");
+          }
```
jschwartz-cray: This approach is fine by me instead of #177.  Thanks for including this to fix that issue.

PR #185, opened 2019-09-09
Add DAOS based drivers (DAOS Array & DFS)

mchaarawi:
Add 2 drivers based on DAOS (the Distributed Asynchronous Object Store).

DAOS is an open source software-defined scale-out object store that provides high-bandwidth, low latency, and high-IOPS storage containers. DAOS is architected from the ground up to exploit new NVM technologies and is extremely lightweight as it operates end-to-end in user space.
https://github.com/daos-stack/daos

The 2 Drivers added with this PR are both based on top of the DAOS API:
1. The Native DAOS driver using the DAOS object array API, which maps a file to a DAOS 1 dimensional byte array object.
2. The DFS driver which based on top of the DFS library which export an encapsulated namespace with a POSIX-like API to users (traditional files and directories). Underneath the hood, DFS does the mapping from files and directories to native DAOS objects. (https://github.com/daos-stack/daos/blob/master/src/client/dfs/README.md)

Both drivers support IOR, but only DFS supports mdtest.


src/mdtest-main.c
```diff
@@ -2,12 +2,9 @@
 #include "aiori.h"
 
 int main(int argc, char **argv) {
-    aiori_initialize(NULL);
```
JulianKunkel: Great patch. Could you clarify why you removed aiori init and finalize from main?

src/option.c
```diff
@@ -308,11 +316,14 @@ static void option_parse_token(char ** argv, int * flag_parsed_next, int * requi
                 if(strlen(arg) > 1){
                   printf("Error, ignoring remainder of string for option %c (%s).\n", o->shortVar, o->longVar);
                 }
-                break;
-              }
+	      }
```
JulianKunkel: just checking why you removed the break here?

mchaarawi: hmm that look accidental, will fix..

src/mdtest-main.c
mchaarawi: if i remember right, the init and finalize calls, at least for the DFS driver, require input from the user (pool, container uuids) and those are not done until we parse the command line options in mdtest_run().
please let me know if you have a better suggestion and i can update.

PR #187, opened 2019-09-10
IOR: Add latency and IOPS numbers to each iteration.

mchaarawi:
IOR currently reports IOPS in the summary but doesn't report it per iteration. This patch adds IOPS numbers for each iteration in the output. IO Latency is also important to look at, especially for small transfer sizes. We report that too for each iteration.

- Latency reported is computed by taking the average latency of
  all ops from a single task, then taking the minimum of that
  between all tasks.

- IOPS is computed by taking the total number of ops across all
  tasks divided by the total access time to execute those ops.

### DIFF
--- a/doc/USER_GUIDE
+++ b/doc/USER_GUIDE
@@ -550,6 +550,17 @@ HOW DOES IOR CALCULATE PERFORMANCE?
   operations (-g), the sum of the open, transfer, and close times may not equal
   the elapsed time from the first open to the last close.
 
+  After each iteration (-i) IOR reports performance for that iteration, and
+  those numbers include:
+
+  - Bandwidth (described above)
+
+  - IOPS: I/O rate (operations per second) achieved by all tasks given the total
+    time spent in reading and writing the data.
+
+  - Latency: computed by taking the average latency of all I/O operations from a
+    single task. If ior is run with multiple tasks, then the latency reported is
+    the minimum that was computed between all tasks.
 
 HOW DO I ACCESS MULTIPLE FILE SYSTEMS IN IOR?
 
--- a/src/aiori-DAOS.c
+++ b/src/aiori-DAOS.c
@@ -241,7 +241,6 @@ DAOS_Init()
         if (rank == 0) {
                 uuid_t			uuid;
 		d_rank_list_t		*svcl = NULL;
-                d_rank_list_t    	ranks;
 		static daos_pool_info_t po_info;
 		static daos_cont_info_t co_info;
 
@@ -368,6 +367,8 @@ DAOS_Create(char *testFileName, IOR_param_t *param)
 	/** Distribute the array handle if not FPP */
 	if (!param->filePerProc)
 		HandleDistribute(&aoh, ARRAY_HANDLE);
+
+	return &aoh;
 }
 
 static int
@@ -417,6 +418,8 @@ DAOS_Open(char *testFileName, IOR_param_t *param)
 	/** Distribute the array handle if not FPP */
 	if (!param->filePerProc)
 		HandleDistribute(&aoh, ARRAY_HANDLE);
+
+	return &aoh;
 }
 
 static IOR_offset_t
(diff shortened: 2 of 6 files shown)


mchaarawi:
> Looks alright to me.

Hi Julian, Is this OK to land?

JulianKunkel:
yeah good to me, but let have another one looking.

mchaarawi:
> Looks good to me. Except ... why take the minimum reported latency across all ranks? That seems like it will give an artificially inflated score.

I don't think that would matter much since i'd thing for latency measurements, you would only use 1 task. And for that we compute the average of all transfers from that task. So that will not be an inflated score.
When you have multiple tasks, my opinion is that the minimum is appropriate between the tasks themselves, but TBH if we take max or average, i don't think it would matter much since it makes more sense to get latency with a single task. Anyway that is just my opinion, and i don't mind to change the results to report average for example.

johnbent:
As long as it is documented what is being reported, then I'm fine with
this.  Thanks for explanation.

jti-lanl:
For parallel operations, wouldn't we want the actual end-to-end latency, from launching the tasks through the final barrier?

That would capture the true “cost of doing business” in the parallel regime.

A more-nuanced approach would report the average per-task latency, in addition to the end-to-end latency, so we could separate the cost of parallel control from the benefit of parallel streams.



On Sep 16, 2019, at 9:41 AM, Mohamad Chaarawi <notifications@github.com<mailto:notifications@github.com>> wrote:


Looks good to me. Except ... why take the minimum reported latency across all ranks? That seems like it will give an artificially inflated score.

I don't think that would matter much since i'd thing for latency measurements, you would only use 1 task. And for that we compute the average of all transfers from that task. So that will not be an inflated score.
When you have multiple tasks, my opinion is that the minimum is appropriate between the tasks themselves, but TBH if we take max or average, i don't think it would matter much since it makes more sense to get latency with a single task. Anyway that is just my opinion, and i don't mind to change the results to report average for example.

glennklockwood:
I don't think there's a right answer @jti-lanl.  Sometimes you want to measure application performance, sometimes you want to measure absolute system capability.  I tend to agree with @johnbent; best thing to do is just document the semantics of this metric.

I don't have time to look at the code more closely, but might there be a way to use `-g` (`intraTestBarriers`) or similar to allow users to choose which definition of latency they want?

mchaarawi:
> For parallel operations, wouldn't we want the actual end-to-end latency, from launching the tasks through the final barrier?

The idea behind this metric was to just measure the latency of an IO operation with a particular transfer size, which I think is what IOR is really about (raw data performance of read/writes). 
IOR does report the cost of open/close, but i really don't see why even those should be included in the average latency we are reporting, since per iteration you do 1 open, 1 close, and all IOPS. Plus if someone is really interested in IOPS and latency of metadata operations, mdtest would be better in this case to use vs IOR.

But if i understood right, what you are suggesting is measuring the latency of an entire workflow? That is a different metric which can be reported at the end of a run, and not per-iteration. so that would be a different patch IMO.

jti-lanl:
Okay, thanks.  Sorry for the interruption.


On Sep 16, 2019, at 11:51 AM, Mohamad Chaarawi <notifications@github.com<mailto:notifications@github.com>> wrote:


For parallel operations, wouldn't we want the actual end-to-end latency, from launching the tasks through the final barrier?

The idea behind this metric was to just measure the latency of an IO operation with a particular transfer size, which I think is what IOR is really about (raw data performance of read/writes).
IOR does report the cost of open/close, but i really don't see why even those should be included in the average latency we are reporting, since per iteration you do 1 open, 1 close, and all IOPS. Plus if someone is really interested in IOPS and latency of metadata operations, mdtest would be better in this case to use vs IOR.

But if i understood right, what you are suggesting is measuring the latency of an entire workflow? That is a different metric which can be reported at the end of a run, and not per-iteration. so that would be a different patch IMO.

mchaarawi:
I updated the user guide to indicate what those new metrics added correspond too.
The code already had some comments in there to explain how we compute those.

johannlombardi:
is this patch ready to be landed then?

JulianKunkel:
I'd be OK with this, surely needs some testing which can be achieved for the next release version. As the IOR developers wanted to follow the 4-eye  principles, someone else should give the OK / or directly merge: @glennklockwood @johnbent

PR #188, opened 2019-09-12
Enable global default dir layout for subdirs in Lustre

pkoutoupis:
that is, delete inherited striping layout with the invocation of a command line parameter during command execution. Feature is made available only when running configure and building with the "with-lustre" option.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -91,7 +91,8 @@ AS_IF([test "x$with_lustre" = xyes ], [
                         AC_MSG_FAILURE([--with-lustre was given, <lustre/lustre_user.h> not found])
                 fi
         ])
-        AC_CHECK_HEADERS([linux/lustre/lustreapi.h lustre/lustreapi.h], break, [
+        AC_CHECK_HEADERS([linux/lustre/lustreapi.h lustre/lustreapi.h],
+		[AC_DEFINE([HAVE_LUSTRE_LUSTREAPI], [], [Lustre user API available in some shape or form])], [
                 if test "x$with_lustre" != xcheck -a \
                         "x$ac_cv_header_linux_lustre_lustreapi_h" = "xno" -a \
                         "x$ac_cv_header_lustre_lustreapi_h" = "xno" ; then
--- a/configure.ac
+++ b/configure.ac
@@ -83,7 +83,7 @@ AC_ARG_WITH([lustre],
         [AS_HELP_STRING([--with-lustre],
                 [support configurable Lustre striping values @<:@default=check@:>@])],
         [], [with_lustre=check])
-AS_IF([test "x$with_lustre" != xno], [
+AS_IF([test "x$with_lustre" = xyes ], [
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -83,14 +83,25 @@ AC_ARG_WITH([lustre],
         [AS_HELP_STRING([--with-lustre],
                 [support configurable Lustre striping values @<:@default=check@:>@])],
         [], [with_lustre=check])
-AS_IF([test "x$with_lustre" != xno], [
```
glennklockwood: What was the motivation behind this change?  It looks like `--with-gpfs` right above is still using the `!= xno`.

configure.ac
```diff
@@ -83,14 +83,25 @@ AC_ARG_WITH([lustre],
...
+AS_IF([test "x$with_lustre" = xyes ], [
         AC_CHECK_HEADERS([linux/lustre/lustre_user.h lustre/lustre_user.h], break, [
                 if test "x$with_lustre" != xcheck -a \
                         "x$ac_cv_header_linux_lustre_lustre_user_h" = "xno" -a \
                         "x$ac_cv_header_lustre_lustre_user_h" = "xno" ; then
                         AC_MSG_FAILURE([--with-lustre was given, <lustre/lustre_user.h> not found])
                 fi
         ])
+        AC_CHECK_HEADERS([linux/lustre/lustreapi.h lustre/lustreapi.h], break, [
```
glennklockwood: This exposes the same bug as #189 so we'll have to rebase that fix on this patch.

src/mdtest.c
```diff
@@ -76,6 +76,10 @@
 
 #include <mpi.h>
 
+#ifdef HAVE_LUSTRE_LUSTREAPI_H
```
glennklockwood: This will break if `linux/lustre/lustre_user.h` is the path to `lustre_user.h` (see #189)

configure.ac
pkoutoupis: Everything that isn't explicitly set to "yes" or (= xyes) is **!= xno**. Even if garbage data is somehow pulled in. I was experiencing issues earlier on in my testing because of this, when I was **not** building --with-lustre and it was still falling into that same code path. To maintain form, the others can be modified as part of this or another PR. Or if you prefer, I can revert while to discuss our options.

configure.ac
pkoutoupis: I will rebase as soon as #190 (which fixes #189) is merged to master. As of now, I see that it is not.

src/mdtest.c
pkoutoupis: Please see note above.

configure.ac
pkoutoupis: @glennklockwood: any additional thoughts on this?

configure.ac
pkoutoupis: I guess I do not have to wait until #190 is pushed. I can make the changes and push before that.

PR #196, opened 2019-10-28
update DFS driver with new DAOS API changes.

mchaarawi:


glennklockwood:
Should there be ifdefs checking for DAOS versions in here?  Or is nobody realistically going to use the old API?

mchaarawi:
> Should there be ifdefs checking for DAOS versions in here? Or is nobody realistically going to use the old API?

Yes we would need this when we have 1.0 released. For now considering where we are, it should be OK to just require be using latest DAOS master.

PR #199, opened 2019-11-05
PMDK backend with dual mount options

adrianjhpc:
This is fork with a new backend for PMDK functionality that will write directly to NVRAM DIMMs. The PMDK backend is reasonably straight forward (quite similar to the POSIX one) but the positioning of the persist functions may need checking (I think they are properly mapped to equivalent fsync functionality but I could be wrong).

The slight complexity is to get good performance on a dual socket node with NVRAM using PMDK it's necessary to have dual mount points is that processes on different sockets use a different file (and thereby a different mount point). To support this, there's a new option (dualMount) which should be only coded to work with the individual file case (PMDK only supports this at the moment).

This dualMount option is a bit tricky with the provided filename, because now it expects the testFile name to point to a directory which it will be parse and add a number and filename to, i.e.L

 testFile=/mnt/pmem_fsdax

code parses this, adds 0 or 1 depending on what socket the process is running on, and a filename, to become something like:

/mnt/pmem_fsdax0/data with the file names individual processes would use normally after that
likewise /mnt/pmem_fsdax1/data

Not entirely pleasant functionality, but until PMDK provides NUMA aware allocators its necessary.

Let me know if there any questions


JulianKunkel:
Nice commit Adrian,
however, the CI does fail, please fix the autotool build chain to make it work!

adrianjhpc:
Done, at least for the minimal build.

JulianKunkel:
Thanks Adrian, note that with the changed module structure the PMDK AIORI module would need an update. Its options can be integrated into the module as well. Albeit these changes would be needed to make it operational, I found the patch good to have.

PR #200, opened 2019-11-07
DAOS driver update

mchaarawi:
- Update configure check for DAOS to check for libs under daos_dir/lib64
- Add DFS option for the dfs prefix

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -206,7 +206,7 @@ AC_ARG_WITH([daos],
 
 AS_IF([test "x$with_daos" != xno], [
     DAOS="yes"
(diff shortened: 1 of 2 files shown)


JulianKunkel:
That looks OK to me. I believe as long as the DAOS API is not stable, we should merge such changes with lower frequency.

mchaarawi:
> That looks OK to me. I believe as long as the DAOS API is not stable, we should merge such changes with lower frequency.

yea these kind of things should stabilize more as we draw near 1.0. Till then, there might still be minor changes to the build and API that we would want to push upstream.
After 1.0 we will ensure ABI/API backward compatibility of the DAOS drivers.

PR #201, opened 2019-11-11
Not sure why iops and latency are left out of the JSON job output. It…

rldleblanc:
… is really useful information.

Signed-off by: Robert LeBlanc <robert@leblancnet.us>


JulianKunkel:
They were added later and not included in the JSON output. Thx for adding.

PR #208, opened 2019-12-16
Fix confusing output of mknod

jyvet:
Context: mknod does not return a file descriptor.


glennklockwood:
Yikes, thanks for fixing this.

PR #215, opened 2020-02-26
DAOS Driver: Update License & Align with 1.0 API changes

mchaarawi:
- License should inherit from IOR License.
- remove checksums from Array API
- add DFS dir oclass option

### DIFF
--- a/src/aiori-DAOS.c
+++ b/src/aiori-DAOS.c
@@ -1,16 +1,9 @@
-/*
- * -*- mode: c; c-basic-offset: 8; indent-tabs-mode: nil; -*-
+/* -*- mode: c; c-basic-offset: 8; indent-tabs-mode: nil; -*-
  * vim:expandtab:shiftwidth=8:tabstop=8:
  */
 /*
- * Copyright (C) 2018-2019 Intel Corporation
- *
- * GOVERNMENT LICENSE RIGHTS-OPEN SOURCE SOFTWARE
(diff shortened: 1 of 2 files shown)


JulianKunkel:
I'm having a little concern regarding the change of the license of an industry contribution that  removes some rights previously granted. In the strict sense, Copyright:
"means whatever content the author created cannot be used or published by anyone else without the consent of the author.". Hence, could mean that Intel needs to agree before a TAR of this code is published somewhere. 
I would welcome it if you'd be able to add a disclaimer underneath that allows the code to retain the GPL2 license which the general IOR code is.
I also ask others for their feedback before merging.




I do not see a problem with any change to actual code...

mchaarawi:
> I'm having a little concern regarding the change of the license of an industry contribution that removes some rights previously granted. In the strict sense, Copyright:
> "means whatever content the author created cannot be used or published by anyone else without the consent of the author.". Hence, could mean that Intel needs to agree before a TAR of this code is published somewhere.
> I would welcome it if you'd be able to add a disclaimer underneath that allows the code to retain the GPL2 license which the general IOR code is.
> I also ask others for their feedback before merging.
> 
> I do not see a problem with any change to actual code...

yes that was definitely not my intention. I made a mistake pushing under APL, since APL is not viral, but GPL is.
I'll add the disclaimer to retain the GPL2 license.

mchaarawi:
I reproduced what other commercial contributors had under their driver. please let me know if there is still a concern.

PR #219, opened 2020-03-17
Improve offsets-generation and uniformity offset count for the N-1 case.

pkoutoupis:
Patch to improve the efficiency of the offsets-generation and uniformity of offset count for the N-1 case.

This patch addresses the following:

1.  IOR generates an array for the sequence of offsets.  The GetOffsetArraySequential() and GetOffsetArrayRandom() function calls are included in the timed test.  For the Random offsets, this can potentially be a significant overhead with very small offsets to very large files and impact the reported performance.  Both of the GetOffsetArray functions are rewritten to improve efficiency.

2.  For the random offset array generated for the single-shared-file (N-1) case, this includes a random number of offsets for processes such that the load balance is not uniform.  With the change to GetOffsetArrayRandom(), the single-shared-file case now has a uniform number of offsets per rank.

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -1719,7 +1719,7 @@ static void ValidateTests(IOR_param_t * test)
         }
 }
 
-/**
+/*
  * Returns a precomputed array of IOR_offset_t for the inner benchmark loop.
  * They are sequential and the last element is set to -1 as end marker.
  * @param test IOR_param_t for getting transferSize, blocksize and SegmentCount
@@ -1744,33 +1744,50 @@ static IOR_offset_t *GetOffsetArraySequential(IOR_param_t * test,
         offsetArray[offsets] = -1;      /* set last offset with -1 */
 
         /* fill with offsets */
-        for (i = 0; i < test->segmentCount; i++) {
-                for (j = 0; j < (test->blockSize / test->transferSize); j++) {
-                        offsetArray[k] = j * test->transferSize;
-                        if (test->filePerProc) {
-                                offsetArray[k] += i * test->blockSize;
-                        } else {
-                                offsetArray[k] +=
-                                        (i * test->numTasks * test->blockSize)
-                                        + (pretendRank * test->blockSize);
+        IOR_offset_t next_offset, incrament, trips;
+        /* special cases for filePerProc and transferSize=blockSize */
+        if (test->filePerProc) {
+                next_offset = 0;
+                incrament = test->transferSize;
+                trips = test->segmentCount * (test->blockSize / test->transferSize);
+                for (i = 0; i < trips; i++) {
+                        offsetArray[i] = next_offset;
+                        next_offset += incrament;
+                }
+        } else {
+                if (test->transferSize == test->blockSize) {
+                        next_offset = pretendRank * test->blockSize;
+                        incrament = test->numTasks * test->blockSize;
+                        trips = test->segmentCount;
+                        for (i = 0; i < trips; i++) {
+                                offsetArray[i] = next_offset;
+                                next_offset += incrament;
                         }
-                        k++;
+                } else {        /* usual case is segmentCount=1 */
+                        next_offset = pretendRank * test->blockSize;
+                        incrament = test->transferSize;
+                        trips = test->blockSize / test->transferSize;
+                        for (i = 0; i < test->segmentCount; i++) {
+                                for (j = 0; j < trips; j++) {
+                                        offsetArray[k+j] = next_offset;
+                                        next_offset += incrament;
+                                }
+                                k += trips;
+                                next_offset += (test->numTasks * test->blockSize);
+                        }
                 }
         }
 
         return (offsetArray);
 }
 
-/**
+/*
  * Returns a precomputed array of IOR_offset_t for the inner benchmark loop.
  * They get created sequentially and mixed up in the end. The last array element
  * is set to -1 as end marker.
  * It should be noted that as the seeds get synchronised across all processes
- * every process computes the same random order if used with filePerProc.
- * For a shared file all transfers get randomly assigned to ranks. The processes
- * can also have differen't numbers of transfers. This might lead to a bigger
- * diversion in accesse as it dose with filePerProc. This is expected but
- * should be mined.
+ * every process computes the same random order.  The processes * also have the
+ * same number of transfers so load balance is uniform, if run * to completion.
  * @param test IOR_param_t for getting transferSize, blocksize and SegmentCount
  * @param pretendRank int pretended Rank for shifting the offsest corectly
  * @return IOR_offset_t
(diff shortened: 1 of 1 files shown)


src/ior.c
```diff
@@ -1744,33 +1744,50 @@ static IOR_offset_t *GetOffsetArraySequential(IOR_param_t * test,
...
+        /* special cases for filePerProc and transferSize=blockSize */
+        if (test->filePerProc) {
+                next_offset = 0;
+                incrament = test->transferSize;
+                trips = test->segmentCount * (test->blockSize / test->transferSize);
+                for (i = 0; i < trips; i++) {
+                        offsetArray[i] = next_offset;
+                        next_offset += incrament;
+                }
+        } else {
```
johnbent, line 1757: Maybe add a '/* single shared file */' comment after '} else {'.

src/ior.c
```diff
@@ -1744,33 +1744,50 @@ static IOR_offset_t *GetOffsetArraySequential(IOR_param_t * test,
...
+                if (test->transferSize == test->blockSize) {
+                        next_offset = pretendRank * test->blockSize;
+                        incrament = test->numTasks * test->blockSize;
+                        trips = test->segmentCount;
+                        for (i = 0; i < trips; i++) {
+                                offsetArray[i] = next_offset;
+                                next_offset += incrament;
+                        }
+                } else {        /* usual case is segmentCount=1 */
```
johnbent, line 1766: Do we need an explicit test for segmentCount==1?  What happens here if it isn't?

src/ior.c
```diff
@@ -1794,53 +1811,39 @@ static IOR_offset_t *GetOffsetArrayRandom(IOR_param_t * test, int pretendRank,
...
-                        if ((rand() % test->numTasks) == pretendRank) {
-                                offsetArray[offsetCnt] = i;
-                                offsetCnt++;
-                        }
-                }
-        }
         /* reorder array */
-        for (i = 0; i < offsets; i++) {
-                value = rand() % offsets;
+        if (offsets < 2*first) first = offsets;
```
johnbent, line 1821: What is special about the number 2 here?

src/ior.c
```diff
@@ -1794,53 +1811,39 @@ static IOR_offset_t *GetOffsetArrayRandom(IOR_param_t * test, int pretendRank,
...
-                        // a shared file
-                        if ((rand() % test->numTasks) == pretendRank) {
-                                offsets++;
-                        }
-                } else {
-                        offsets++;
-                }
-        }
+        /* start with same array of offsets as in sequential case */
+        offsetArray = GetOffsetArraySequential(test, pretendRank);
```
johnbent, line 1818: Here we are in the random branch?  But we start with sequential?  Does this make it less random than it used to be?

src/ior.c
```diff
@@ -1794,53 +1811,39 @@ static IOR_offset_t *GetOffsetArrayRandom(IOR_param_t * test, int pretendRank,
...
+                for (j = 0; j < vecs; j++) {
+                        start[j] = i + 8*(random() % (offsets - i - vlen)/8);  // align on a cache line
+                }
+                seq = 0;
+                for (k = 0; k < vlen; k++) {
+                        for (j = 0; j < vecs; j++) {
+                                value = start[j];
+                                tmp = offsetArray[value+k];
+                                offsetArray[value+k] = offsetArray[i+seq];
+                                offsetArray[i+seq] = tmp;
```
johnbent, line 1841: I wonder if the three previous lines would look a bit cleaner with a helper function:
swapArrayElements(offsetArray,value+k,i+seq);
I think this would look better and be slightly easier to read.

src/ior.c
```diff
@@ -1744,33 +1744,50 @@ static IOR_offset_t *GetOffsetArraySequential(IOR_param_t * test,
...
-        for (i = 0; i < test->segmentCount; i++) {
-                for (j = 0; j < (test->blockSize / test->transferSize); j++) {
-                        offsetArray[k] = j * test->transferSize;
-                        if (test->filePerProc) {
-                                offsetArray[k] += i * test->blockSize;
-                        } else {
-                                offsetArray[k] +=
-                                        (i * test->numTasks * test->blockSize)
-                                        + (pretendRank * test->blockSize);
+        IOR_offset_t next_offset, incrament, trips;
```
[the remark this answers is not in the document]
pkoutoupis: Yup. Not sure what I was thinking. Must have been a bad substitute. Thanks for catching it.

pkoutoupis:
@JulianKunkel Is there a document or format that I can reference for a unit test?

glennklockwood:
We've struggled to decide on a proper testing framework, and at present, it is limited to what's in `testing/*-tests.sh` which essentially runs a full (but small) IOR test.

I am not enough of a computer scientist to know a reasonable way to test randomness in the offset generation subroutine.  Hopefully someone else can help.  Worst case, `-vvv` can be passed to dump exact offsets during a `-z` run and we have a postprocessor in the `testing/` directory that asserts that the random offsets are random enough...maybe?

JulianKunkel:
I had started the testing under src/test when the API was created. I have simplified adding a test now and added one for the "sequential". Should hopefully be easier to add more tests now.

JulianKunkel:
To collect some more evidence, I have done some benchmarking on my laptop regarding the current strategy in order to reconsider.  At the moment, all the offsets are precomputed (even for sequential access), this increases the memory footprint and the runtime. When using stonewalling with very high numbers it may cause performance issues.
Test ran with 1,000,000,000 segments. Takes about 7.5 GByte memory (8 Byte per offset).
With sequential, takes 3.5s to precompute. 
Random (-z), 183s which seems significant.
This means 5,470,676 offsets are computed per second. Even for 4096-byte accesses, the file would be 3.73 TByte large for such a testcase.

IMHO it would be best to integrate the calculation of the offsets into the processing instead of pre-computing them. That would avoid the memory consumption and the overall need.
For random, though, this would imply that no full shuffling of all segments can be done anymore. Similar to this patch, I consider that it is likely OK to have a pattern with a number of segments defined by the user (e.g., 1 million) that are shuffled, and then repeat this pattern over and over again. That would still result in random access over a range of 4 GByte which seems to me OK. I'll inclined to test such a patchset.

JulianKunkel:
I have now implemented the new behavior in the PR listed above and done quite some testing.
To make a note about performance: 
- there is now no storage needed for sequential and no precompute.
- the example I gave above: with 1,000,000,000 segments, could be rewritten to for example use a blocksize of 1/100th of the segments listed, these the time to pre-calculate offsets (needed only 1x as before) would be ~1.8s, the area that is truly random would then be 1/100th of the overall file size, e.g. 37 GByte out of the 3.73 TByte. I personally consider that to be random enough (otherwise, could use bigger block size) which then though needs more time and memory to pre-calculate/store these offsets.

JulianKunkel:
I'm closing this in favor of: Fix offset integration #270

PR #224, opened 2020-05-29
src/aiori-CEPHFS: Option warnings, shutdown checks, and cmount nullification

markhpc:
Fixes io500 C application invocation of the aiori-CEPHFS backend so long as the API is set in the individual benchmark sections rather than globally.  It also throws a WARN if Init is called before the options are initialized (which is what IO500 tries to do currently if you set the API globally in the ini file).


JulianKunkel:
Great patch!

PR #225, opened 2020-05-30
Improve AIORI module behavior

JulianKunkel:
I took the opportunity to start a bigger refactoring project.
The branch referenced by the pull request aims to reduce the dependency of the AIORI interface to the ior_test_param_t structure, and to clarify/simplify the AIORI interface.
Basically, the goal is to not depend on any IOR specific parameter as this is confusing and misleading in tools such as mdtest. The AIORI interface could be generally useful to provide a low-level abstraction layer to IO for IO benchmarks.
I do not intend that we accept the pull request now but instead use it as a discussion point of issues and gather comments.

The analysis shows that various modules depend too much on IOR options and make assumptions about the access pattern when a file is opened. I believe this should be fixed in the long run, for now a hint function is provided to AIORI:
void init_xfer_options(IOR_param_t * params);
If this is called, it basically is a hint to the backend that specific access patterns will take place. In the long run it should be replaced with a proper hint structure about the access pattern instead of passing all information and should be optional. 
By keeping such a structure, it makes the usage of IOR options more visible in the first step.

The patch also moves various module-specific options of the POSIX module and HDF5 module into the respective modules to cleanup the ior_param structure.
The BeeGFS/Lustre/GPFS specific options need testing/inspection by further stakeholders.

A roadmap to reduce the dependency and clarify the situation would be:
1) Get an agreement of the semantics of what AIORI should do (and this strategy), at the moment too many dependencies to IOR specifics such as param->offset which each module deals *individually*. IMHO that should be dealt in calls to AIORI.
2) Invite stakeholders of the respective modules as reviewers to create pull requests to the branch to fix the modules to the current "back-compatibility" level. This will lead to a stabilization of options and their documentation which is useful to document the changed behavior.
3) Define the "hint" structure for expected patterns which may be exploited in various AIORI calls.
4) Modify the calls to AIORI and the module implementations.


JulianKunkel:
> it's really weird to me that the offset of where the transfer has to happen is passed as a "hint" and not actually part of the function xyz_xfer.

I absolutely agree to this. It was the behavior of IOR and before in the central IOR structure together with everything, I changed as little as necessary yet enough to achieve a first clarification of the API and separation of IOR from AIORI. The offset part can be fixed in a second iteration as it would change the AIORI interface in non-trivial way and may lead to some compatibility issues with modules. 
For now, would you agree that this patchset generally is a big step in the right direction or is there anything else that should be worked on?

mchaarawi:
> > it's really weird to me that the offset of where the transfer has to happen is passed as a "hint" and not actually part of the function xyz_xfer.
> 
> I absolutely agree to this. It was the behavior of IOR and before in the central IOR structure together with everything, I changed as little as necessary yet enough to achieve a first clarification of the API and separation of IOR from AIORI. The offset part can be fixed in a second iteration as it would change the AIORI interface in non-trivial way and may lead to some compatibility issues with modules.
> For now, would you agree that this patchset generally is a big step in the right direction or is there anything else that should be worked on?

no i agree with you that the change is good in separating the aiori functionality from ior and before it was weird with mdtest. so yes not objections on that part.
i was just saying that it might be confusing as is to backend developers that they would need to get required params from a global hint struct. glad you agree this is an interim patch and a future one would address that.

mchaarawi:
OK it was more work than anticipated, but i did fix the DAOS and DFS modules to work with this new model. created a PR to this branch:
https://github.com/hpc/ior/pull/229
I probably will have follow on commits to improve some things, but i can do that after all this goes to master.

In the short future, i do want to "merge" both the DAOS and DFS drivers into a single driver. It would be more like remove the DAOS driver and rename DFS to DAOS.

mchaarawi:
I see some conflicts, i will fix them

PR #227, opened 2020-06-02
mdtest: allocate aligned buffers to support DirectIO

breuner:
`mdtest --posix.odirect` was not working and failed with "Invalid argument" on write/read (e.g. on RHEL7 for xfs/BeeGFS/NFS), because the corresponding memory buffers for write/read were not correctly aligned. According to `man 2 open`, "The O_DIRECT flag may impose alignment restrictions on the length and address of user-space buffers". This patch fixes DirectIO support by allocating page-aligned write and read buffers via `posix_memalign`.

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -633,8 +633,8 @@ void mdtest_read(int random, int dirs, const long dir_iter, char *path) {
 
     /* allocate read buffer */
     if (read_bytes > 0) {
-        read_buffer = (char *)malloc(read_bytes);
-        if (read_buffer == NULL) {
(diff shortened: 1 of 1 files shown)


breuner:
Just noticed a copy&paste mistake: Currently using `write_bytes` for the read buffer allocation, which should actually be `read_bytes`. Will you fix it directly or would you like me to send a new PR?

glennklockwood:
If you push a patch to your branch I think we can re-use this pull request.  If not, no big deal to do another one.

breuner:
Fix pushed to my branch, but does not look like this PR got updated, so I'll open a new one.

PR #229, opened 2020-06-11
update DFS and DAOS drivers to new aiori module interface.

mchaarawi:

### DIFF
--- a/src/aiori-DAOS.c
+++ b/src/aiori-DAOS.c
@@ -24,15 +24,16 @@
 #include <libgen.h>
 #include <stdbool.h>
 
+#include <mpi.h>
 #include <gurt/common.h>
 #include <daos.h>
 
-#include "ior.h"
 #include "aiori.h"
+#include "utilities.h"
 #include "iordef.h"
 
 /************************** O P T I O N S *****************************/
-struct daos_options{
+typedef struct {
         char		*pool;
         char		*svcl;
         char		*group;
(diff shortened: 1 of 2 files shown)


src/aiori-DFS.c
```diff
@@ -27,15 +27,15 @@
...
 #include <daos_fs.h>
 
-#include "ior.h"
-#include "iordef.h"
 #include "aiori.h"
 #include "utilities.h"
+#include "iordef.h"
 
 dfs_t *dfs;
 static daos_handle_t poh, coh;
```
JulianKunkel, line 41: I'm a bit concerned about the use of several global daos handles, instead of returning them as a pointer. 
Does this support a proper semantics opening multiple "files/objects" at a time?

mchaarawi: yes those are the pool & container handles.
The pool is the storage unit (controls capacity, storage reservation).
The container is the object address space. In POSIX world, think of it as a namespace.
so for IOR and mdtest, the expectation is always to have a single pool, and access to a single namespace (daos container / dfs mount).
A file in DFS is represented as an object (dfs_obj_t - see what is returned in DFS_Create). and from this model you can have billions/trillions of objects (files, directories) per namespace.

hope that is clear. for more info please see:
https://daos-stack.github.io/overview/storage/

src/aiori-DAOS.c
```diff
@@ -334,38 +362,38 @@ DAOS_Fini()
...
 		DCHECK(rc, "Failed to create array object\n");
 	}
 
 	/** Distribute the array handle if not FPP */
-	if (!param->filePerProc)
+	if (!hints->filePerProc)
 		HandleDistribute(&aoh, ARRAY_HANDLE);
 
-	return &aoh;
+	return (aiori_fd_t*)(&aoh);
```
JulianKunkel, line 392: Here is the line that fits together with my concern.

mchaarawi: yes, you are referring to the DAOS array driver.
This is a legacy implementation that was done using the DAOS Array API (before the DFS layer was implemented). As I mentioned, the DFS driver is the way to go, and i will soon push a patch to remove this entire driver.

JulianKunkel:
That's great! Now I understand that the new API fixes this.

It is true that at the moment each of these benchmarks opens only one
file at a time but there was (and hopefully will be) a version that
run two benchmarks at the same time.

Btw. one could generally move static globals into the
"module-specific" structure. That would allow to initiate a backend
multiple times (with different backends). Maybe this can be done
during the cleanup as well.
Thanks.

PR #236, opened 2020-07-02
aiori-IME: Update to new aiori interface

jyvet:
This patch updates IME backend to support new aiori interface.
It also fixes some indentation issues.

### DIFF
--- a/src/aiori-IME.c
+++ b/src/aiori-IME.c
@@ -21,8 +21,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/stat.h>
-#include <errno.h>                                  /* sys_errlist */
-#include <fcntl.h>                                  /* IO operations */
+#include <errno.h>                              /* sys_errlist */
+#include <fcntl.h>                              /* IO operations */
 
 #include "ior.h"
 #include "iordef.h"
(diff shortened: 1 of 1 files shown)


src/aiori-IME.c
```diff
@@ -100,92 +105,113 @@ ior_aiori_t ime_aiori = {
...
 
 /*
  * Try to access a file through the IME interface.
  */
-static int IME_Access(const char *path, int mode, IOR_param_t *param)
+
+int IME_Access(const char *path, int mode, aiori_mod_opt_t *module_options)
 {
-        (void)param;
+        (void)module_options;
```
JulianKunkel, line 170: Note: this is for silencing compiler warnings/errors.

src/aiori-IME.c
```diff
@@ -371,28 +399,32 @@ static int IME_RmDir(const char *path, IOR_param_t *param)
...
-        if (ime_native_stat(testFileName, &stat_buf) != 0) {
-                ERR("cannot get status of written file");
-        }
+        if (ime_native_stat(testFileName, &stat_buf) != 0)
+                ERRF("cannot get status of written file %s", testFileName);
+
         aggFileSizeFromStat = stat_buf.st_size;
 
-        if (test->filePerProc) {
+        if (hints->filePerProc) {
```
JulianKunkel, line 427: This code part can be removed.

src/aiori-IME.c
jyvet: Yes, I'll add a comment here.

src/aiori-IME.c
jyvet: thanks, I'm going to fix that.

jyvet:
Many thanks for the review! I'll update the path shortly.

src/aiori-IME.c
jyvet: Seems this is not done yet. I'm getting warnings if I remove the reduction.

```
WARNING: Expected aggregate file size       = 17179869184.
WARNING: Stat() of aggregate file size      = 4294967296.
WARNING: Using actual aggregate bytes moved = 17179869184.
```

JulianKunkel: Could you specify which options you have used?
The code responsible is in ior.c:327 and following. That was used in POSIX before and I wonder how that could have happened here.

jyvet: It does not work with FPP as in ior.c:327 the reduction only concerns aggFileSizeFromXfer. With FPP it can only be compared to aggFileSizeFromStat if there was a previous reduction on all files before. For now this is done in each aiori backend.

jyvet:
I've simplified the `IME_GetFileSize` function. It seems that the reduction cannot be removed yet for FFP.

src/aiori-IME.c
JulianKunkel: Now I see the confusion: could you check with the branch: feature-cleanup-aiori
That one contains some further changes regarding the file size.
I merge this pull now, the delta to the other branch should be minimal.

JulianKunkel:
Please submit a pull request to the branch feature-cleanup-aiori

Issue #249, opened 2020-08-11
IOR '-R' flag increases memory pressures and causes performance impacts


sihara:
If -R option in IOR enables, it causes big memory pressures on data verification for read.
Here is a quick profiling result of memory bandwidth during IOR with/without -R flag. It was tested on a Intel client (1 x Gold 5218, 1 x IB-HDR100). And ior reproted 12.5GB/sec in this csae regardless with/without '-R'.
The memory bandwidth standpoint, without '-R' flag, it was 14GB/s Write and 2.8GB/s Read, but with '-R' flag, It consumed 30GB/s Read and 28GB/s Write.

```codes
Without '-R' : mpirun -np 16 ior -r -a POSIX -t 16m --posix.odirect -b 70000m \
-F -i 1 -C -Q 1 -g -G 27 -k -e -o /es400nv/ior_file_easy -O stoneWallingStatusFile=/es400nv/stonewall -D 100
|---------------------------------------|
|--             Socket  0             --|
|---------------------------------------|
|--     Memory Channel Monitoring     --|
|---------------------------------------|
|-- Mem Ch  0: Reads (MB/s):   506.94 --|
|--            Writes(MB/s):  2348.85 --|
|-- Mem Ch  1: Reads (MB/s):   493.25 --|
|--            Writes(MB/s):  2348.26 --|
|-- Mem Ch  2: Reads (MB/s):   514.79 --|
|--            Writes(MB/s):  2348.90 --|
|-- Mem Ch  3: Reads (MB/s):   450.81 --|
|--            Writes(MB/s):  2325.86 --|
|-- Mem Ch  4: Reads (MB/s):   462.17 --|
|--            Writes(MB/s):  2326.62 --|
|-- Mem Ch  5: Reads (MB/s):   444.50 --|
|--            Writes(MB/s):  2325.87 --|
|-- NODE 0 Mem Read (MB/s) :  2872.45 --|
|-- NODE 0 Mem Write(MB/s) : 14024.36 --|
|-- NODE 0 P. Write (T/s):      18751 --|
|-- NODE 0 Memory (MB/s):    16896.81 --|
|---------------------------------------|
|---------------------------------------||---------------------------------------|
|--                 System Read Throughput(MB/s):       2872.45                --|
|--                System Write Throughput(MB/s):      14024.36                --|
|--               System Memory Throughput(MB/s):      16896.81                --|
|---------------------------------------||---------------------------------------|

[...]

PR #254, opened 2020-09-02
Integrated comparison checks into CompareData() which prevents additi…

JulianKunkel:
Change buffer comparison behavior to minimize memory pressure: #249

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -374,36 +374,55 @@ static void CheckFileSize(IOR_test_t *test, char * testFilename, IOR_offset_t da
  * difference in buffers and returns total errors counted.
  */
 static size_t
-CompareBuffers(void *expectedBuffer,
-               void *unknownBuffer,
-               size_t size,
-               IOR_offset_t transferCount, IOR_param_t *test, int access)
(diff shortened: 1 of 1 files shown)


adilger (approved):

adilger (approved):

src/ior.c
```diff
@@ -374,36 +374,56 @@ static void CheckFileSize(IOR_test_t *test, char * testFilename, IOR_offset_t da
...
 
         length = size / sizeof(IOR_size_t);
-        first = -1;
         if (verbose >= VERBOSE_3) {
                 fprintf(out_logfile,
                         "[%d] At file byte offset %lld, comparing %llu-byte transfer\n",
                         rank, (long long) offset, (long long)size);
         }
+
+        int incompressibleSeed = test->setTimeStampSignature + fillrank;
```
johnbent, line 409: This line does not seem intuitive at all.

adilger: This is by necessity the same as used elsewhere in the code to generate the PRNG data (e.g. line 645).

PR #262, opened 2020-10-26
aiori-Gfarm: update to the new aiori interface

otatebe:
This change updates aiori-Gfarm.c to catch up with the new aiori interface.


JulianKunkel:
That looks good, thanks!

PR #270, opened 2020-11-06
Fix offset integration

JulianKunkel:
This series of patches addresses #219.
The behavior of the offset creation had been changed as follows:
- there is no need to create the array of all file offsets any more
- for sequential IO, the offset calculation is part of the loops
- for random IO, the random offset calculation is done as before but only within *one* block providing some meaning to the semantics of block, transfer size, and segments. Before the blocks were meaningless for random.

Now: the segments repeat the same random pattern.

Consider the example for 1 process:
-t 4096 -b 40960 -s 2
There are 10 random positions between offsets 0, 4k, up to 36k, this pattern is repeated two times (offset by 40kb in the second iteration).

When more than 1 process takes place, it depends on if it is file per process or shared file.
For FPP, each process will now have its own random pool for a block, otherwise similar to 1 process case.
For shared file, the randomness is number of processes * block and the overall pattern is randomized in this larger area -- this behavior is as it was before. However, the same pattern repeats then for each segment.

To imitate the old behavior, one can set -s=1 and use a large block size. 
To create for example a random pattern that effects a range of 4 MByte and then repeats 10x, use:
-t 4096 -b 4m -s 10

This allows now to reduce the pressure of storing offsets for random while it still gives the flexibility to control what's happening.

### DIFF
--- a/src/aiori.c
+++ b/src/aiori.c
@@ -128,6 +128,8 @@ void aiori_supported_apis(char * APIs, char * APIs_legacy, enum bench_type type)
 {
         ior_aiori_t **tmp = available_aiori;
         char delimiter = ' ';
+        *APIs = 0;
+        *APIs_legacy = 0;
 
         while (*tmp != NULL)
         {
@@ -136,7 +138,6 @@ void aiori_supported_apis(char * APIs, char * APIs_legacy, enum bench_type type)
                     tmp++;
                     continue;
                 }
-
                 if (delimiter == ' ')
                 {
                         APIs += sprintf(APIs, "%s", (*tmp)->name);
@@ -148,6 +149,7 @@ void aiori_supported_apis(char * APIs, char * APIs_legacy, enum bench_type type)
                 if ((*tmp)->name_legacy != NULL)
                         APIs_legacy += sprintf(APIs_legacy, "%c%s",
                                                delimiter, (*tmp)->name_legacy);
+
                 tmp++;
         }
 }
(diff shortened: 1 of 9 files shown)


glennklockwood:
Changing the semantics of `-b` when `-z` is specified makes me nervous because it means previous IOR configs may no longer be random if used with this newer version of the code.  However bigger picture of making -b have some purpose in random mode is great, and I like the new way of randomizing offsets.

How should we navigate changing the behavior of `-b` and `-z`?  New minor release?  We are already pretty far behind on releases, with 3.3 only being a release candidate and 3.4 (as I recall) having similarly big changes to the core code.

JulianKunkel:
> Changing the semantics of -b when -z is specified makes me nervous
because it means previous IOR configs may no longer be random if used with
this newer version of the code.
I absolutely agree, that's why I integrated a check if -b = -t and then
report an error. While that won't catch all cases, it should catch most of
them.

Glad you like it.

How should we navigate changing the behavior of -b and -z? New minor
> release? We are already pretty far behind on releases, with 3.3 only being
> a release candidate and 3.4 (as I recall) having similarly big changes to
> the core code.
>
That is a tricky question. A minor release we should do. I think it is
actually OK to have 3.4 as RC and integrate md-workbench.
I'll add some documentation about the new features in a minor release then
(basically exposing to the users that the feature is there).

JulianKunkel:
This is now the new behavior for random
image

x-special/nautilus-clipboard
copy
file:///home/kunkel/Downloads/IOR-SSF-random-new.pptx

PR #276, opened 2020-11-24
Update in documentation

JulianKunkel:
This PR is created for the new branch that will improve the documentation.
It isn't finished yet but for discussions reasons, I created the PR immediately.


adilger (commented):
.pptx isn't a good format for storing into git, since it requires MS Office to open.  Using .png would at least be viewable more easily by everyone.

johnbent:
Plantuml and graphviz are excellent because they are text-based which also can be rendered directly in github.com pages:

https://github.com/Seagate/cortx/blob/main/doc/images/graphviz/README.md

And 

https://github.com/Seagate/cortx/blob/main/doc/images/plantuml/README.md

JulianKunkel:
I've opened the document with Libreoffice without a problem.
I agree that SVG would be preferable. I've merely modified a document created by a user and added it as it is IMHO good to have source files in the documentation with it, where as alternative of not having any. As there are not too many contributions to documentation, I value all of them highly.

PR #281, opened 2020-11-26
Bugfix MDTest calculation of multiple iterations was incorrect.

JulianKunkel:
Fix the bug reported by Rick to increase clarity. Thanks!

The previous offset calculation when using multiple iterations was:
for (i = start; i < stop; i++) // i = table position == test number
  for (k=0; k < size; k++) 
     for (j = 0; j < iterations; j++) 
        value = all[(k \* tableSize \* iterations) + (j*tableSize) + i];

Note that the mean and min/max was then computed over these values.
But as the values were stored in memory in the order:  iteration, rank, table
the correct term is: value = all[j \* tableSize \* size + k * tableSize + i];

Assume iterations = 2 and size = 3, the value for the test i=0 was computed from:
all[0 * 2 *tbl + 0 * tbl] = 0tbl
all[0 * 2 *tbl + 1 * tbl] = 1tbl
all[1 * 2 *tbl + 0 * tbl] = 2tbl
all[1 * 2 *tbl + 1 * tbl] = 3tbl
all[2 * 2 *tbl + 0 * tbl] = 4tbl
all[2 * 2 *tbl + 1 * tbl] = 5tbl

A more clear traversal would have been:
all[0 * 3 *tbl + 0 * tbl] = 0tbl
all[0 * 3 *tbl + 1 * tbl] = 1tbl
all[0 * 3 *tbl + 2 * tbl] = 2tbl
all[1 * 3 *tbl + 0 * tbl] = 3tbl
all[1 * 3 *tbl + 1 * tbl] = 4tbl
all[1 * 3 *tbl + 2 * tbl] = 5tbl

In that sense, it wasn't a functional bug but it decreased readability and now that we want to print the performance of the individual ranks, it is useful to fix this.

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -1291,6 +1291,20 @@ void summarize_results(int iterations, int print_time) {
       }
     }
 
+    if(print_all_proc && 0){
+      // This code prints the result table for debugging
+      for (i = 0; i < tableSize; i++) {
+        for (j = 0; j < iterations; j++) {
+          access = mdtest_test_name(i);
+          if(access == NULL){
+            continue;
+          }
+          curr = summary_table[j].rate[i];
+          fprintf(out_logfile, "Rank %d Iter %d Test %s Rate: %e\n", rank, j, access, curr);
+        }
+      }
+    }
+
     if (rank != 0) {
       return;
     }
@@ -1314,7 +1328,6 @@ void summarize_results(int iterations, int print_time) {
         start = stop = 0;
     }
 
-
     if(print_all_proc){
       fprintf(out_logfile, "\nPer process result (%s):\n", print_time ? "time" : "rate");
       for (j = 0; j < iterations; j++) {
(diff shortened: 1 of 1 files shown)


src/mdtest.c
```diff
@@ -1340,10 +1353,9 @@ void summarize_results(int iterations, int print_time) {
 
     for (i = start; i < stop; i++) {
             min = max = all[i];
-            for (k=0; k < size; k++) {
-                for (j = 0; j < iterations; j++) {
-                    curr = all[(k*tableSize*iterations)
-                               + (j*tableSize) + i];
+            for (j = 0; j < iterations; j++) {
```
adilger, line 1356: Using better variable names than "i" and "j" and "k", like "iter" and "op" might avoid this kind of bug in the future.

Also, putting the "j*tableSize*size + k*tableSize + i" calculation into a small helper function like:

    int calc_allreduce_index(int index, int iter, int op)

(or whatever) and using it in the places where all[] is accessed would also avoid similar usage bugs.

adilger:
> The previous offset calculation when using multiple iterations was:

This kind of information would be very useful in the commit message of the patch itself, rather than just the PR, since it will be easily available with the code in the future.

PR #283, opened 2020-11-27
Refactoring: MDTest

JulianKunkel:
This PR refactors the 50+ static variables to be part of a global option structure to remove ambiguity in which scope the variable is.

In the code, there is at the moment too much dependency to global variables.
Ultimately, the goal was to make stonewalling work with the multi-dir approach but it turned out this requires much more refactoring and therefore, doing it piece by piece.

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -149,7 +149,6 @@ typedef struct {
   int pre_delay;
   int unique_dir_per_task;
   int time_unique_dir_overhead;
-  int throttle;
   int collective_creates;
   size_t write_bytes;
   int stone_wall_timer_seconds;
@@ -1885,7 +1884,6 @@ void mdtest_init_args(){
    memset(& o, 0, sizeof(o));
    o.barriers = 1;
    o.branch_factor = 1;
-   o.throttle = 1;
 }
 
 mdtest_results_t * mdtest_run(int argc, char **argv, MPI_Comm world_com, FILE * world_out) {
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -90,87 +90,92 @@
 
 #define LLU "%lu"
 
-static int size;
-static uint64_t *rand_array;
-static char testdir[MAX_PATHLEN];
-static char testdirpath[MAX_PATHLEN];
-static char base_tree_name[MAX_PATHLEN];
-static char **filenames;
-static char hostname[MAX_PATHLEN];
-static char mk_name[MAX_PATHLEN];
-static char stat_name[MAX_PATHLEN];
-static char read_name[MAX_PATHLEN];
-static char rm_name[MAX_PATHLEN];
-static char unique_mk_dir[MAX_PATHLEN];
-static char unique_chdir_dir[MAX_PATHLEN];
-static char unique_stat_dir[MAX_PATHLEN];
-static char unique_read_dir[MAX_PATHLEN];
-static char unique_rm_dir[MAX_PATHLEN];
-static char unique_rm_uni_dir[MAX_PATHLEN];
-static char *write_buffer;
-static char *stoneWallingStatusFile;
+typedef struct {
+  int size;
+  uint64_t *rand_array;
+  char testdir[MAX_PATHLEN];
+  char testdirpath[MAX_PATHLEN];
+  char base_tree_name[MAX_PATHLEN];
+  char **filenames;
+  char hostname[MAX_PATHLEN];
+  char mk_name[MAX_PATHLEN];
+  char stat_name[MAX_PATHLEN];
+  char read_name[MAX_PATHLEN];
+  char rm_name[MAX_PATHLEN];
+  char unique_mk_dir[MAX_PATHLEN];
+  char unique_chdir_dir[MAX_PATHLEN];
+  char unique_stat_dir[MAX_PATHLEN];
+  char unique_read_dir[MAX_PATHLEN];
+  char unique_rm_dir[MAX_PATHLEN];
+  char unique_rm_uni_dir[MAX_PATHLEN];
+  char *write_buffer;
+  char *stoneWallingStatusFile;
 
 
-static int barriers;
-static int create_only;
-static int stat_only;
-static int read_only;
-static int verify_read;
-static int verify_write;
-static int verification_error;
-static int remove_only;
-static int leaf_only;
-static unsigned branch_factor;
-static int depth;
+  int barriers;
+  int create_only;
+  int stat_only;
+  int read_only;
+  int verify_read;
+  int verify_write;
+  int verification_error;
+  int remove_only;
+  int leaf_only;
+  unsigned branch_factor;
+  int depth;
 
-/*
- * This is likely a small value, but it's sometimes computed by
- * branch_factor^(depth+1), so we'll make it a larger variable,
- * just in case.
- */
-static uint64_t num_dirs_in_tree;
-/*
- * As we start moving towards Exascale, we could have billions
- * of files in a directory. Make room for that possibility with
- * a larger variable.
- */
-static uint64_t items;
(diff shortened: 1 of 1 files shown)


src/mdtest.c
```diff
@@ -1021,25 +1026,77 @@ int updateStoneWallIterations(int iteration, uint64_t items_done, double tstart,
...
-    VERBOSE(0,-1, "Continue stonewall hit min: %lld max: %lld avg: %.1f \n", min_accessed, max_iter, ((double) sum_accessed) / size);
+  if(o.items != (sum_accessed / o.size)){
+    VERBOSE(0,-1, "Continue stonewall hit min: %lld max: %lld avg: %.1f \n", min_accessed, max_iter, ((double) sum_accessed) / o.size);
     hit = 1;
   }
 
   return hit;
 }
 
+void file_test_create(const int iteration, const int ntasks, const char *path, rank_progress_t * progress, double *t){
```
adilger, line 1047: It would be better to move this code to its own function in a different patch.  This patch is already changing several hundred lines to use the options structure, and adding other changes into the patch makes it harder to review.

src/mdtest.c
```diff
@@ -1021,25 +1026,77 @@ int updateStoneWallIterations(int iteration, uint64_t items_done, double tstart,
...
+  char temp_path[MAX_PATHLEN];
+  int cur_dir_loops = o.directory_loops;
```
adilger, line 1049: (style) it isn't clear what the benefit of "cur_dir_loops" is over using "o.directory_loops" directly on the next line?

src/mdtest.c
```diff
@@ -1021,25 +1026,77 @@ int updateStoneWallIterations(int iteration, uint64_t items_done, double tstart,
...
+        // now reset the values
+        progress->stone_wall_timer_seconds = o.stone_wall_timer_seconds;
+        o.items = progress->items_done;
+      }
+      if (o.stoneWallingStatusFile){
+        StoreStoneWallingIterations(o.stoneWallingStatusFile, max_iter);
+      }
+      // reset stone wall timer to allow proper cleanup
+      progress->stone_wall_timer_seconds = 0;
+      break;
```
adilger, line 1094: This "break" is not in the original code, and looks like it is changing the control flow.  It is hard to say for sure since there is no explanation in the commit message of why it is here.  It *looks* like a bug that would cause the loop to *always* exit on the first iteration? 

This kind of subtle change shouldn't be hidden in a patch with hundreds of other lines of mechanical changes, as it is very hard to spot.

src/mdtest.c
```diff
@@ -1871,44 +1880,10 @@ static void mdtest_iteration(int i, int j, MPI_Group testgroup, mdtest_results_t
...
-   read_bytes = 0;
-   sync_file = 0;
-   call_sync = 0;
-   path_count = 0;
-   nstride = 0;
-   make_node = 0;
-#ifdef HAVE_LUSTRE_LUSTREAPI
-   global_dir_layout = 0;
-#endif /* HAVE_LUSTRE_LUSTREAPI */
+   memset(& o, 0, sizeof(o));
```
adilger, line 1884: Using a struct initializer will set all of the fields to zero except those specified:

    o = {
       .barriers = 1,
       .branch_factor = 1,
       .throttle = 1,
    };

src/mdtest.c
```diff
@@ -2161,122 +2136,122 @@ mdtest_results_t * mdtest_run(int argc, char **argv, MPI_Comm world_com, FILE *
...
+    for (i = first; i <= last && i <= o.size; i += stride) {
         range.last = i - 1;
         MPI_Group_range_incl(worldgroup, 1, (void *)&range, &testgroup);
         MPI_Comm_create(testComm, testgroup, &testComm);
         if (rank == 0) {
-            uint64_t items_all = i * items;
-            if(num_dirs_in_tree_calc){
-              items_all *= num_dirs_in_tree_calc;
+            uint64_t items_all = i * o.items;
+            if(o.num_dirs_in_tree_calc){
```
adilger, line 2242: The normal style in the rest of the code is "if (" and ") {} (with a space in between).  Minor whitespace cleanups like this could also be done in this patch.

src/mdtest.c
```diff
@@ -2286,39 +2261,39 @@ mdtest_results_t * mdtest_run(int argc, char **argv, MPI_Comm world_com, FILE *
...
     }
 
-    if (created_root_dir && remove_only && backend->rmdir(testdirpath, backend_options) != 0) {
-        FAIL("Unable to remove test directory path %s", testdirpath);
+    if (created_root_dir && o.remove_only && o.backend->rmdir(o.testdirpath, o.backend_options) != 0) {
+        FAIL("Unable to remove test directory path %s", o.testdirpath);
     }
 
-    if(verification_error){
+    if(o.verification_error){
```
adilger, line 2281: (style) "if (" and ") {"

src/mdtest.c
```diff
@@ -373,65 +378,65 @@ static void create_file (const char *path, uint64_t itemNum) {
...
          * According to Bill Loewe, writes are only done one time, so they are always at
          * offset 0 (zero).
          */
-        hints.fsyncPerWrite = sync_file;
-        if(write_bytes >= 8){ // set the item number as first element of the buffer to be as much unique as possible
-          ((uint64_t*) write_buffer)[0] = itemNum;
+        o.hints.fsyncPerWrite = o.sync_file;
+        if(o.write_bytes >= 8){ // set the item number as first element of the buffer to be as much unique as possible
+          ((uint64_t*) o.write_buffer)[0] = itemNum;
         }else{
```
adilger, line 421: (style) "} else {"

src/mdtest.c
```diff
@@ -373,65 +378,65 @@ static void create_file (const char *path, uint64_t itemNum) {
...
 
         /*
          * According to Bill Loewe, writes are only done one time, so they are always at
          * offset 0 (zero).
          */
-        hints.fsyncPerWrite = sync_file;
-        if(write_bytes >= 8){ // set the item number as first element of the buffer to be as much unique as possible
-          ((uint64_t*) write_buffer)[0] = itemNum;
+        o.hints.fsyncPerWrite = o.sync_file;
+        if(o.write_bytes >= 8){ // set the item number as first element of the buffer to be as much unique as possible
```
adilger, line 419: (style) "if (" and ") {"

adilger (changes requested):
It looks like the added "break" may have broken the code?

src/mdtest.c
JulianKunkel: I see where you come from, maybe time to thing about some coding conventions. In terms of what is used in the repo though:
$ grep if src/*.c| cut -d ":" -f 2- |grep "if([a-zA-Z0-9]" | wc -l
264
$ grep "if ( " src/*.c| cut -d ":" -f 2- |grep "if([a-zA-Z0-9]" | wc -l
70
In that sense, the style I used is dominant 80%. I personally also think this isn't that important (right now) but yeah, maybe coding conventions.

src/mdtest.c
JulianKunkel: I agree, will do it next time. It was part of the attempt to refactor the "stonewall" which then lead to the situation that I refactored the "static" pieces instead to increase understandability.

src/mdtest.c
JulianKunkel: At the moment, the stonewalling feature works only with 1 directory, there is a check to ensure that the loop iterates only 1, so it doesn't change the behavior. But well spotted. Part of the first patch that aimed to refactor this part of the code. Yeah, agreed the refactored version must have more comments!

src/mdtest.c
JulianKunkel: Very good suggestion, made.

PR #292, opened 2020-12-03
Do not execute functions twice in MPI_CHECKF

ofaaland:
Assigning MPI_STATUS to a local variable and then referring to the
local will ensure that the same value is used in both the conditional
expression and the call to MPI_Error_string.

Otherwise, when MPI_STATUS is a function call, like

  MPI_CHECKF(fubar(), "%s", "error in fubar");

fubar() is called twice.  If there are underlying intermittent errors,
the error code/message for the first call is lost, with confusing output
like this:

  read      2206.18    17.27      145.93      262144     131072
  0.272595   291.88     0.290829   292.41
  ERROR: cannot access explicit, noncollective, MPI MPI_SUCCESS: no errors, (aiori-MPIIO.c:451)

### DIFF
--- a/src/aiori-debug.h
+++ b/src/aiori-debug.h
@@ -94,12 +94,13 @@ extern int aiori_warning_as_errors;
 #define MPI_CHECKF(MPI_STATUS, FORMAT, ...) do {                         \
     char resultString[MPI_MAX_ERROR_STRING];                             \
     int resultLength;                                                    \
+    int checkf_mpi_status = MPI_STATUS;                                  \
                                                                          \
-    if (MPI_STATUS != MPI_SUCCESS) {                                     \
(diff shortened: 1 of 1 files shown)


JulianKunkel (approved):

JulianKunkel:
Good catch! Will merge it once the regression test passed.

ofaaland:
I'm getting type-related warnings on the build I overlooked initially, so I'll get those quashed and then push a revision.

ofaaland:
OK, I found the bug in my patch, it's fixed.  I was using "status" as the local variable name, which conflicted with a variable name used in the function call, e.g.
```
                                MPI_CHECK(Access_all
                                          (mfd->fd, buffer, length,
                                           mfd->transferType, &status),
                                          "cannot access collective");
```

PR #295, opened 2020-12-11
added region and location to libs3

Frankgad:
This PR gives the user the ability to specify the region and the location of the bucket.
This is useful when testing against public cloud provider. More Info can be found here:
https://docs.aws.amazon.com/AmazonS3/latest/dev/optimizing-performance-guidelines.html#optimizing-performance-guidelines-combine

### DIFF
--- a/src/aiori-S3-libs3.c
+++ b/src/aiori-S3-libs3.c
@@ -56,6 +56,7 @@ static option_help * S3_options(aiori_mod_opt_t ** init_backend_options, aiori_m
 
   *init_backend_options = (aiori_mod_opt_t*) o;
   o->bucket_prefix = "ior";
+  o->bucket_prefix_cur = "b";
 
   option_help h [] = {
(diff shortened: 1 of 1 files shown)


src/aiori-S3-libs3.c
```diff
@@ -116,7 +119,7 @@ static void def_bucket_name(s3_options_t * o, char * out_name, char const * path
   }
   *out_name = '\0';
 
-  // S3Status S3_validate_bucket_name(const char *bucketName, S3UriStyle uriStyle);
+  S3Status S3_validate_bucket_name(const char *bucketName, S3UriStyle uriStyle);
```
JulianKunkel, line 122: The function isn't called here. So it is just a declaration. Can you comment it out or use it?

Frankgad: you're right! Since the different S3 providers impose different rules on the length of bucket names and their structure, i did comment it out at this stage.

PR #299, opened 2020-12-19
fix #281 ("MDTest calculation of multiple iterations was incorrect") for 3.3 release branch

### LINKED ISSUE #298: backport #281 into 3.3 branch
the fix in #281 should be backported to the 3.3 branch

glennklockwood:
A reproducer wasn't included with #281 so I haven't tested to make sure this is complete, but the patch in #281 seemed pretty straightforward.

This should resolve #298.

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -1231,7 +1231,7 @@ void file_test(const int iteration, const int ntasks, const char *path, rank_pro
 
 int calc_allreduce_index(int iter, int rank, int op){
   int tableSize = MDTEST_LAST_NUM;
-  return iter * tableSize * o.size + rank * tableSize + op;
+  return iter * tableSize * size + rank * tableSize + op;
 }
 
 void summarize_results(int iterations, int print_time) {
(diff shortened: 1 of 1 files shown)


glennklockwood:
Also if/when we merge this, can we **squash+merge**?  I noticed a couple of pull requests weren't squashed before merging which have polluted our commit history.

JulianKunkel (commented):
On the new Line 1301, need to replace all[....] with all[calc_allreduce_index(j, k, i)] as well. Otherwise, looks good.

glennklockwood:
> On the new Line 1301, need to replace all[....] with all[calc_allreduce_index(j, k, i)] as well. Otherwise, looks good.

I was wondering about that.  Does this mean that the following patch (on the 3.2 branch) should also be changed?

https://github.com/hpc/ior/blob/7a490c87b544ab7bae588b4cf3de43c67e033536/src/mdtest.c#L1510

Without a reproducer I can't easily tell what the effects are.

glennklockwood:
> Without a reproducer I can't easily tell what the effects are (also because I'm too lazy to refamiliarize myself with the mdtest code).

Nevermind.  Unless someone objects, I'll leave 3.2 alone.

Issue #304, opened 2020-12-24
make distcheck fails


glennklockwood:
`make distcheck` fails on src/test/example.c for a couple reasons.

On master,

```
/Applications/Xcode.app/Contents/Developer/usr/bin/make  testlib testexample
  CC       lib.o
  CCLD     testlib
  CC       example.o
../../../../src/test/example.c:3:10: fatal error: 'ior.h' file not found
#include <ior.h>
         ^~~~~~~
1 error generated.
make[4]: *** [example.o] Error 1
make[3]: *** [check-am] Error 2
make[2]: *** [check-recursive] Error 1
make[1]: *** [check-recursive] Error 1
make: *** [distcheck] Error 1
```

This is easily resolved by changing `<ior.h>` to `"../ior.h"` as src/test/lib.c does.  The next error is

```
/Applications/Xcode.app/Contents/Developer/usr/bin/make  testlib testexample
  CC       lib.o
  CCLD     testlib
  CC       example.o
../../../../src/test/example.c:20:13: error: implicit declaration of function 'GetOffsetArraySequential' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  offsets = GetOffsetArraySequential(& test, 0);
            ^
../../../../src/test/example.c:20:11: warning: incompatible integer to pointer conversion assigning to 'IOR_offset_t *' (aka 'long long *') from 'int' [-Wint-conversion]
  offsets = GetOffsetArraySequential(& test, 0);
          ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning and 1 error generated.
make[4]: *** [example.o] Error 1
make[3]: *** [check-am] Error 2
make[2]: *** [check-recursive] Error 1
make[1]: *** [check-recursive] Error 1
make: *** [distcheck] Error 1
```

`GetOffsetArraySequential` was taken out of ior.c in 3ee3e9a, but example.c was never updated.

JulianKunkel:
Can you check the version in the PR?

glennklockwood:
Note for next release: there is a known issue where `make distcheck` will not work if build is performed using an MPICH-based MPI stack due to dependency on MPI in the runtime environment.  OpenMPI will work fine.

PR #305, opened 2020-12-25
Updated test (illustration of an empty test) and distclean for make d…

JulianKunkel:
…istcheck #304


glennklockwood:
Thanks!  This works for OpenMPI which is good enough for me.

MPICH builds fail because MPICH needs PMI for `MPI_Init()` to succeed, but that's going to introduce way too much complexity to this seeing as how we currently don't use it.

PR #309, opened 2021-01-18
use rank 0 for the work dir creation/rmdir.

a-lyashkov:
fix an mdtest start/cleanup issue when several
work directories point to the same location via
diffent mount points.
Number a work dirs isn't large so it have no differences
with old style.

Signed-off-by: Alexey Lyashkov <a.lyashkov@yandex.com>

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -1705,7 +1705,7 @@ static void mdtest_iteration(int i, int j, MPI_Group testgroup, mdtest_results_t
     prep_testdir(j, dir_iter);
 
     VERBOSE(2,5,"main (for j loop): making o.testdir, '%s'", o.testdir );
-    if ((rank < o.path_count) && o.backend->access(o.testdir, F_OK, o.backend_options) != 0) {
+    if ((rank == 0) && o.backend->access(o.testdir, F_OK, o.backend_options) != 0) {
         if (o.backend->mkdir(o.testdir, DIRMODE, o.backend_options) != 0) {
             FAIL("Unable to create test directory %s", o.testdir);
         }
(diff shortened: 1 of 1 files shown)


JulianKunkel:
Hi, good catch. I wonder how that could live that long in the tree.
I made a slightly a more complete implementation and extracted the rank.
Can you confirm that solves the issue #310

JulianKunkel:
I see now, the path_count was intended to support that different MPI ranks can have different paths.
Something like @dir1@dir2@dir3
I reckon, this feature hasn't been used and isn't working anymore properly for a while now, so I would remove the stubs of that.

glennklockwood:
I've used `@dir1@dir2@dir3` to test Lustre remote directory performance in aggregate, and it can be used to test multi-mounted NFS.  

I would like to keep that feature if at all possible, but I admit that I don't know when the last time I verified that it still works is.

JulianKunkel:
I see, so the use case was, e.g., to run on a single node with 3 procs, and each proc runs on one NFS mounted directory (correct me if this is incorrect). 
I would prefer to have each process access the files round-robin across different directories. That would allow to run with arbitrary many procs / mount points getting rid of some awkward checks.

glennklockwood:
Yes, I think we're describing the same thing.  In concrete terms, I've done two things:

1. Test clustered NFS, where you need multiple mount points to the same nfs backend to get the bandwidth of all the NFS frontends.  All mountpoints have an identical namespace
2. Test Lustre DNE1, where you have to point mdtest at different output directories to measure aggregate performance of all MDSes

I agree that round-robining would reduce complexity, but it changes the nature of what's being tested and may result in different performance than the traditional functionality.  I'm all for reducing complexity; we just need to make sure we note the change in behavior when we first release such a change.

JulianKunkel:
Alright, I think the new PR in #319 now resolves the issue (in my testing) while it keeps the functionality.

PR #310, opened 2021-01-18
MDTest only proc0 to temper with the global dir tree. #309

JulianKunkel:
Please track the discussion under #309


JulianKunkel:
Based on the feedback, we keep the feature.

PR #320, opened 2021-01-21
Remove MPI timer in favor of gettimeofday() to prevent MPI timing issues. 

JulianKunkel:
Remove time adjustment as measurements are relative anyway.

### DIFF
--- a/src/ior-output.c
+++ b/src/ior-output.c
@@ -304,23 +304,6 @@ void PrintHeader(int argc, char **argv)
                 }
                 PrintKeyValEnd();
         }
-
-#ifdef _NO_MPI_TIMER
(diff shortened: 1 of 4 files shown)


glennklockwood:
Can you elaborate on the "MPI issues" you envisioned here?  We depend on MPI so I wouldn't have thought that there would be issues.

JulianKunkel:
Yes, I had been debugging over several days an issue where the timer didn't work properly. 
Sometimes, it returned a value such as 634333 albeit the runtime was 100s.
I traced that back to the timer which used MPI. I don't know exactly why it happens on the system with OpenMPI2.0.X. As we are using other system times for timestamping any way, I explored this further, turned out once I simplified that code it worked.

PR #324, opened 2021-01-29
aiori-DFS: revert change to use testComm in init/fini of DFS driver

mchaarawi:
The DFS driver init and fini phases use MPI_COMM_WORLD to distribute the pool and container handle.
testComm cannot be used there since that is initialized in the io phase not the init phase.

### DIFF
--- a/src/aiori-DFS.c
+++ b/src/aiori-DFS.c
@@ -255,7 +255,7 @@ HandleDistribute(enum handleType type)
                 DCHECK(rc, "Failed to get global handle size");
         }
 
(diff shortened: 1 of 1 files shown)


JulianKunkel:
I suppose, then testComm needs to be initialized earlier. Using MPI_COMM world will break the logic where only a subset of processes take part in an IO test.

JulianKunkel (changes requested):
Can you check if #325 fixes the issue for you.
MDTest appears to were already correct

PR #325, opened 2021-01-29
Move initialization of TestComm before aiori->initialize()

JulianKunkel:
Allows an aiori initialize function to use testComm


mchaarawi (changes requested):
we still need to do the same thing for finalize - move the MPI_Comm_free from the IO function to the finalize.

JulianKunkel:
Alright, yes. I added another patch.

mchaarawi (approved):
thanks! this fixes it

PR #333, opened 2021-02-08
WIP: a histogram of operation times

roblatham00:
I don't think this should be merged as-is yet but I wanted to start a discussion about how to display this information.

Sometimes its helpful to know if a straggler is responsible for slow reported rates.  I could calculate variance and standard deviation but that requires math.  here's a histogram.  blammo you can see outliers right away, or see that everything is pretty close together.

IOR timings look for "first starer" and  "last finisher" so there is presently no way to get this information unless you do what I did and  operate on the timing[] array. 

Actually there is already a histogram in the code, but only used in highly verbose reads.  ( file_hits_histogram ) .  It's been in the tree for over a decade and I don't think I have ever executed that path...

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -843,6 +843,58 @@ static char *PrependDir(IOR_param_t * test, char *rootDir)
 
 /******************************************************************************/
 /*
+ * Dump histogram of operation times from every process
+ */
+static void
+PrintHistogram(IOR_test_t *test, double *timer, const int rep, const int access)
+{
+        double access_time, min, max, step;
+        int i, nprocs;
+        int bins[] = {0, 0, 0, 0, 0, 0};
+        int nr_bins = sizeof(bins)/sizeof(bins[0]);
+
+        /* do these indicies have names I could use? */
+        access_time = timer[3] - timer[2];
+        /* i am most interested in seeing how tight the write times are.  I
+         * expect most of the times to be quite a bit larger than zero.  I can
+         * see an argument for starting the bins at zero though... */
+
+        double *all_times;
+        MPI_Comm_size(testComm, &nprocs);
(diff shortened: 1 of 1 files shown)


JulianKunkel:
I think this is a very important feature to have. I would opt to store the information into a CSV file, e.g., one row per rank that contains then the details about the measured performance of this specific rank and allow the user to decide on the external tools to render different graphs such as histograms. 
Later, that could be extended easily to include 5-point statistics (or more) about individual reads of each Rank itself. MD-Workbench, for instance, reports the statistics for each metadata op type.

JulianKunkel:
I've submitted a patch to implement storing the individual performance numbers to a CSV file.
I like your comment regarding the timer numbers, indeed that should be refactored using ENUMs or such.

Example output of the CSV file when running
$ mpiexec -np 2  ./src/ior -O saveRankPerformanceDetailsCSV=test.csv
is 
access,rank,runtime-with-openclose,runtime,throughput-withopenclose,throughput
write,0,2.7801990509e-03,2.7048587799e-03,3.7715860656e+08,3.8766386171e+08
write,1,3.1809806824e-03,3.1049251556e-03,3.2963922284e+08,3.3771377648e+08
read,0,3.4618377686e-04,3.2901763916e-04,3.0289576523e+09,3.1869902254e+09
read,1,3.5905838013e-04,3.3998489380e-04,2.9203496090e+09,3.0841840891e+09

Throughput is measured in Bytes/s.

JulianKunkel:
Rob, if you like, can add another option such as -O outputHistogram and merge your code into it.

JulianKunkel:
Hi Rob, do you still like the feature or is the CSV file fine?

roblatham00:
The CSV is wonderful.  I withdraw this draft.

PR #335, opened 2021-02-10
MDTest allow storing information per rank for later analysis 

JulianKunkel:
when using the --saveRankPerformanceDetails=<FILE> option.

The CSV isn't perfectly formatted, yet, contains a bit much information.


JulianKunkel:
Will need a bit more polishment to ensure that the time without barrier is covered.

Issue #337, opened 2021-02-19
Issue with detection of option with empty value


jyvet:
Options with values are not properly parsed. If the value is not set (say "backend.option=" ), a bug makes appear that 0 was provided.

PR #338, opened 2021-02-19
Fix detection of option with empty value

jyvet:
Context: Options with values are not properly parsed.
If the value is not set, a bug makes appear that 0 was
provided.

This patch fixes options parsing (issue #337). An error
is reported if 'backend.option=' argument is provided.

### DIFF
--- a/src/option.c
+++ b/src/option.c
@@ -253,8 +253,10 @@ static void option_parse_token(char ** argv, int * flag_parsed_next, int * requi
   int i = 0;
   if(arg != NULL){
     arg[0] = 0;
-    arg++;
     replaced_equal = 1;
+
+    // Check empty value
+    arg = (arg[1] == 0) ? NULL : arg + 1;
   }
   *flag_parsed_next = 0;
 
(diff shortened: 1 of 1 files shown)


src/option.c
```diff
@@ -296,7 +298,7 @@ static void option_parse_token(char ** argv, int * flag_parsed_next, int * requi
             case (OPTION_OPTIONAL_ARGUMENT):
             case (OPTION_REQUIRED_ARGUMENT):{
               // check if next is an argument
-              if(arg == NULL){
+              if(arg == NULL && replaced_equal != 1){
```
JulianKunkel, line 301: I'm not quite sure why this is necessary? It seems to me the fix in L259 takes care of it. e.g. if it isn't NULL, then the actual value shall be used.

jyvet: This whole code is hard to read. Here is my understanding:

L258 and L259 make the argument empty in case no value was provided after '=' (without those two lines \0 end-of-string char was used as value in case no value is provided to the option). 

The 'if' block starting at L301, fetches next argument for the value in case arg is NULL. We don't want to do that as when replaced_equal == 1 as we want to return an error (handled line L312). Anyway I don't think we want to check next argument if '=' is detected (would '--backend.option= 1' be valid?)

A few examples:
1) --backend.option=        (before: value 0, with the patch: "Error, argument missing for option...")
2) --backend.option=1      (before: value 1, with the patch: value 1)
3) --backend.option= 1     (before: value 0 and error on unknown 1 argument, with the patch: "Error, argument missing for option...")

JulianKunkel: Yes, patch seems good.

PR #341, opened 2021-03-05
Make noncontiguous I/O work again

### LINKED ISSUE #340: What's the deal with strided accesses?
A decade ago, one could ask IOR to generate a noncontiguous (strided) access.  Today, that is not possible (  https://github.com/hpc/ior/blob/main/src/aiori-MPIIO.c#L123 ).   What happened?

roblatham00:
Fixes  #340 

MPI-IO's treatment of noncontiguous is probably a lot different from the other drivers --- I don't even know what noncontiguous S3 access would mean for example.

I don't know why it was disabled a year ago, but this removes the hard error upon requesting it and fixes a memory allocation problem when using it:

    $ mpiexec -np 4 ./src/ior --mpiio.useFileView --mpiio.showHints --mpiio.useStridedDatatype \
        -s 4 -b 1M -t 1M -c -a MPIIO -o testfile1

There is presently no way to know from the output if the benchmark ran noncontig or contig (one could infer if the "useFileView" information was part of the output summary)

### DIFF
--- a/src/aiori-MPIIO.c
+++ b/src/aiori-MPIIO.c
@@ -72,7 +72,7 @@ static option_help * MPIIO_options(aiori_mod_opt_t ** init_backend_options, aior
     {0, "mpiio.hintsFileName","Full name for hints file", OPTION_OPTIONAL_ARGUMENT, 's', & o->hintsFileName},
     {0, "mpiio.showHints",    "Show MPI hints", OPTION_FLAG, 'd', & o->showHints},
     {0, "mpiio.preallocate",   "Preallocate file size", OPTION_FLAG, 'd', & o->preallocate},
-    {0, "mpiio.useStridedDatatype", "put strided access into datatype [not working]", OPTION_FLAG, 'd', & o->useStridedDatatype},
+    {0, "mpiio.useStridedDatatype", "put strided access into datatype", OPTION_FLAG, 'd', & o->useStridedDatatype},
     //{'P', NULL,        "useSharedFilePointer -- use shared file pointer [not working]", OPTION_FLAG, 'd', & params->useSharedFilePointer},
     {0, "mpiio.useFileView",  "Use MPI_File_set_view", OPTION_FLAG, 'd', & o->useFileView},
       LAST_OPTION
@@ -120,8 +120,6 @@ static int MPIIO_check_params(aiori_mod_opt_t * module_options){
         ERR("segment size must be < 2GiB");
   if (param->useSharedFilePointer)
         ERR("shared file pointer not implemented");
-  if (param->useStridedDatatype)
-        ERR("strided datatype not implemented");
   if (param->useStridedDatatype && (hints->blockSize < sizeof(IOR_size_t)
                                       || hints->transferSize <
                                       sizeof(IOR_size_t)))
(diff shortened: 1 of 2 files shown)


adilger (commented):
I thought that this might have been a recent change, but searching back through the change logs it seems that `useStridedDatatype` has been marked "not working" for quite a long time, at least since commit a56d39e6f "Start with the code from IOR-2.10.3" in 2011.

roblatham00:
It must have worked at some point... the I in IOR is for "interleaved" !  (interleaved or random)

JulianKunkel:
Mhh, it appears that this broke the memory allocation and it still doesn't work.

When running:
$ mpiexec -np 2 ./src/ior -s 1 -b 32K -t 32K -a MPIIO --mpiio.useStridedDatatype --mpiio.useFileView  -k

The resulting file is just 32k big, instead of 64k.

More testing: tried with s=1 and s=2.
Using --mpiio.useFileView alone seems to work.

I believe the MPI-IO driver could fix the situation and faking the situation again by 1) just perform the access if the offset = 0, basically in L397:
                        if (param->useStridedDatatype) {
                                if(offset != 0){
                                  return hints->transferSize;
                                }
Also, when creating the derived data type, pointing it to the same memory buffer repeatedly.

I reset the memory allocation in #347 and started working on this. Let's discuss the behavior.

Besides the buffer issue (which was claimed to be fixed), the combination here appears would work:
mpiexec -np 2 ./src/ior -s 2 -b 32K -t 32K -a MPIIO --mpiio.useStridedDatatype 

However, using --mpiio.useStridedDatatype --mpiio.useFileView at the same time leads to the too small files...

mchaarawi:
as we were discussing on slack, MPIIO should not require the ior proper to allocate the entire buffer for all segments, just so it  can construct its file view. this actually breaks the IOR workflow where the segment count passed can be huge.
MPIIO should be able to construct its file view without the memory buffer.

my other note here is that the IO size in each ior xfer is equal to the transfer size. IOR will not give the drivers transfers of size transfer_size x segment_count. I know that MPIIO lets you construct through derived datatypes a segmented memory buffer, but afaiu there is no setting in IOR that gives the entire segmented buffer in one xfer.

roblatham00:
Ok, one can interpret `--mpiio.useFileView --mpiio.useStridedDatatype ` two different ways: 
- describe this entire thing in one big ol datatype
- cut up the operation into pieces

most MPI-IO implementations will be smart enough to detect "contiguous, but i wrapped it in a datatype" as actually contiguous.

mchaarawi:
> Ok, one can interpret `--mpiio.useFileView --mpiio.useStridedDatatype ` two different ways:
> 
> * describe this entire thing in one big ol datatype
> * cut up the operation into pieces
> 
> most MPI-IO implementations will be smart enough to detect "contiguous, but i wrapped it in a datatype" as actually contiguous.

are you talking about the filetype / file view or the memory type? because you can create a large file view to give the MPIIO library a preview of what the access pattern is like, but the intention was never to submit the entire IO in a single call, was it? IOR will always give you IO of a single transfer size (as the name suggests).

roblatham00:
The name suggests?  the "I" originally stood for interleaved.    I don't think AAABBBCCCDDD counts as interleaved, nor do I think  ABCD then ABCD then .... counts as interleaved either.

mchaarawi:
> The name suggests? the "I" originally stood for interleaved. I don't think AAABBBCCCDDD counts as interleaved, nor do I think ABCD then ABCD then .... counts as interleaved either.

oh i was just talking about the transfer size (not the I in IOR) ;-)
as i mentioned, i think it's actually a good thing that you can create an interleaved file view to give as much info to the MPI library. but the concern here is that you changed ior itself to preallocate the memory buffer for the entire run..
anyway. feel free to discuss this more with the actual maintainers. your change has broke a lot of existing workloads.

roblatham00:
ok, right, triage the existing workflows!  i'm ok with reverting this change while I come up with a different approach

JulianKunkel:
Mhh, I somewhat missed some discussion here. @roblatham00 can you check the newly merged code from #347
I believe this now works (depending on the definition of the semantics).

PR #351, opened 2021-03-20
Support prologue and epilogue for benchmark phases

Kawashima-Azumi:
Help text:

> -^=STRING                     call this external command before each phase (excluded from the timing)
> -$=STRING                     call this external command after each phase (included in the timing)

Usage example:

`
mpirun --allow-run-as-root -np 2 mdtest -C -E -T -r -F -d . -z 3 -b 8 -I 128 -L -u -^ 'echo 3 >/proc/sys/vm/drop_caches' -$ 'sync -f .'
`


src/mdtest.c
```diff
@@ -2230,6 +2215,8 @@ mdtest_results_t * mdtest_run(int argc, char **argv, MPI_Comm world_com, FILE *
       {'Y', NULL,        "call the sync command after each phase (included in the timing; note it causes all IO to be flushed from your node)", OPTION_FLAG, 'd', & o.call_sync},
       {'z', NULL,        "depth of hierarchical directory structure", OPTION_OPTIONAL_ARGUMENT, 'd', & o.depth},
       {'Z', NULL,        "print time instead of rate", OPTION_FLAG, 'd', & o.print_time},
+      {'^', NULL,        "call this external command before each phase (excluded from the timing)", OPTION_OPTIONAL_ARGUMENT, 's', & o.prologue},
```
JulianKunkel, line 2218: I would probably use only a long argument such as --run-before-phase and --run-after-phase
Just move "run-before-phase" into the second element of the struct and get rid of ^ and $.

JulianKunkel (changes requested):
Looks like a good patch! Consider to use long options as this allows all tools (in the long run) to share options.

src/mdtest.c
Kawashima-Azumi: You can commit to <https://github.com/Kawashima-Azumi/ior/tree/prologues> to make such changes.

I've intentionally left it writable by ior developers in particular to make such changes.

JulianKunkel (approved):

PR #352, opened 2021-03-23
Mschaara/dfs 3.3

mchaarawi:


glennklockwood:
Thanks for this!  This represents a pretty big feature change in what should be a minor release, but the changes look like they are really limited only to DAOS, and it enables ior-3.3 to work on more contemporary versions of DAOS.  I'll do basic testing to make sure that nothing in autoconf breaks for weird platforms and I'm happy to include this in 3.3.1.

@JulianKunkel do you have an opinion on whether or not we should include support for newer DAOS in a minor release of IOR?

glennklockwood:
@mchaarawi actually made this because I was asking.  I prefer to use stable releases of IOR, but ior-3.3 doesn't build against newer versions of DAOS.

Ironically, I am also the reason there isn't a newer stable release of IOR that includes the newer DAOS support.  Perhaps my focus should be on cutting a feature freeze for 3.4 as I promised back in December 🙂.

mchaarawi:
> I personally do not believe it is useful in the long run to try and backport changes to any file system API back but would rather be more forward-looking such that people use the newer codebase - in that sense, we should have sufficient good releases if people are in doubt.
> @mchaarawi I would like to understand why the backport was made instead of using the newer version?
> 
> Given that there now is apparently a working backport, seems there was enough interest by DAOS users/pressure, why not then have a minor release.

i did that because @glennklockwood was testing with 3.3 and reported version mismatch with recent DAOS version. 
The version checks were added to DFS after 3.3 was tagged, or were just not ported there. this was my attempt to get a PR to allow people to use 3.3 with any version of DAOS.

Issue #345, opened 2021-03-15
stoneWallingStatusFile does not work for write+read tests


glennklockwood:
It looks like defining `stoneWallingStatusFile` causes problems for jobs that do both write-then-read tests because

1. `ReadStoneWallingIterations` is called at the beginning of each read test in `TestIoSys`, but
2. `StoreStoneWallingIterations` is only called at the end of the entire IOR job

So specifying `ior -w -r -O stoneWallingStatusFile=blah` causes a non-existent file to be read as the stonewalling pairs count.  This then wipes out the actual stonewalling pairs count, which then causes the read test to hit an EOF and abort.

What's the right solution to this?

1. blank out `StoneWallingStatus` if a test is not read-only?
2. correct the behavior and write out the stonewalling file after each write iteration instead of at the end of the ior run?

JulianKunkel:
Good observation. I suppose a clear oversight to not consider typical other configurations but only IO500.
The question is how are people supposed to use it?
Is it safe to assume a user either run write, then read phases in one IOR run or in two (first write, then read) using the same settings, possibly a run/configuration file that performs many tests.

If that would yield, then we could store in the stonewallfile the test number with the iteration count and load it for the respective test.

glennklockwood:
I hit this issue because I mistakenly thought it was required to get stonewalling with wearout to work at all.  I think the easiest solution is to limit the scope of `stoneWallingStatusFile` to work with a single write-only test and print a warning (and ignore the option) in other cases.

If we start keeping per-test entries in a stonewallingstatusfile, it might get confusing because a subsequent read test might use slightly different input parameters (e.g., different number of iterations `-i`) and that option becomes even more confusing.  Ultimately, `stoneWallingStatusFile` only makes sense with `-k`, and the behavior of `-k` in a multi-test IOR run is to only preserve the last run's data files, so the behavior of `stoneWallingStatusFile` should reflect that.

PR #359, opened 2021-04-13
IOR: add several sanity checks for stonewalling #345

JulianKunkel:


glennklockwood (approved):
Thanks for putting these in.  Very helpful.

PR #360, opened 2021-04-20
Adding number of tasks to the CSV output

efajardo:
This pull request adds a feature to include the number of tasks in the CSV output file.

### DIFF
--- a/src/ior-output.c
+++ b/src/ior-output.c
@@ -21,7 +21,7 @@ void PrintTableHeader(){
     fprintf(out_resultfile, "access    bw(MiB/s)  IOPS       Latency(s)  block(KiB) xfer(KiB)  open(s)    wr/rd(s)   close(s)   total(s)   iter\n");
     fprintf(out_resultfile, "------    ---------  ----       ----------  ---------- ---------  --------   --------   --------   --------   ----\n");
   }else if(outputFormat == OUTPUT_CSV){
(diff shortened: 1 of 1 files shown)


src/ior-output.c
```diff
@@ -21,7 +21,7 @@ void PrintTableHeader(){
     fprintf(out_resultfile, "access    bw(MiB/s)  IOPS       Latency(s)  block(KiB) xfer(KiB)  open(s)    wr/rd(s)   close(s)   total(s)   iter\n");
     fprintf(out_resultfile, "------    ---------  ----       ----------  ---------- ---------  --------   --------   --------   --------   ----\n");
   }else if(outputFormat == OUTPUT_CSV){
-    fprintf(out_resultfile, "access,bw(MiB/s),IOPS,Latency,block(KiB),xfer(KiB),open(s),wr/rd(s),close(s),total(s),iter\n");
+    fprintf(out_resultfile, "access,bw(MiB/s),IOPS,Latency,block(KiB),xfer(KiB),open(s),wr/rd(s),close(s),total(s),numTasks,iter\n");
```
adilger, line 24: I just noticed this patch landing to the tree.  I think there are two things wrong with it:
- the header line just above here is not updated to have a "Numtasks" line, so "numTasks" will be printed under the "iter" header.
- inserting a value in the middle of the output list instead of the end, any external script/tool that is parsing these values will now get "iter" and "numtasks" mixed up.  That said, it _may_ be that "iter" is treated specially (it does not have a "key" when printed below), so maybe this is not a bug but should at least be highlighted.

efajardo: @adilger this patch only modfies the output for the CSV format not the default to screen output.

PR #380, opened 2021-07-13
mdtest: make sure fileperproc is set for mdtest read

mchaarawi:
When running mdtest create + read, fileperproc is set properly so
the driver known it's not a single shared file. But when mdtest is
running with only read (-E) with a pre-existing dataset, fileperproc
is never set, and driver thinks it's a single shared file and can
do optimization to share the file handle.


JulianKunkel (approved):

JulianKunkel:
Good catch!

PR #394, opened 2021-10-19
ior.c: Allow for block/transfer size < 8 bytes

sydidelot:
There is a restriction where the minimum block/transfer size is
8 bytes for all the underlying APIs. I believe this check is not
required, especially as specific APIs already have their own check
for blockSize < sizeof(IOR_size_t) and transferSize < sizeof(IOR_size_t)
few lines below the removed lines.

This patch removes this restriction and let the underlying API
check for their own constraints. That allows to benchmark storage
systems with block sizes lower than 8 bytes if supported.


JulianKunkel:
I realize, the backend you mention was in a private repository.
If we see no further interest in this issue by others, we should close this around June 2022.

PR #398, opened 2021-11-23
Port CEPHFS module to current AIORI API, add Lazy I/O support, and Manila support.

pllopis:
This MR makes three contributions:

1. Port CEPHFS to current AIORI API, as it was not compiling anymore. (#393be78)
2. Add Lazy I/O support via a new, optional `--cephfs.olazy` flag. (#1e34b3e)
3. Add a remote mount prefix option. This is necessary for multi-tenancy users such as OpenStack Manila, where the root filesystem is not being mounted, but a volume subpath (our remote_prefix option) is mounted instead. (#53eb94f)

Notably, enabling Lazy I/O enables much better performance for single shared file collective I/O (IO500's ior hard mode):

With `--ceph.olazy`:
```
Summary of all tests:
Operation   Max(MiB)   Min(MiB)  Mean(MiB)     StdDev   Max(OPs)   Min(OPs)  Mean(OPs)     StdDev    Mean(s) Stonewall(s) Stonewall(MiB) Test# #Tasks tPN reps fPP reord reordoff reordrand seed segcnt   blksiz    xsize aggs(MiB)   API RefNum
write        3645.52    3645.52    3645.52       0.00    1822.76    1822.76    1822.76       0.00   32.91714         NA            NA     0      6   1    1   0     1        1         0    0  10000  2097152  2097152  120000.0 CEPHFS      0
```

Without:
```
Summary of all tests:                                                                                                             
Operation   Max(MiB)   Min(MiB)  Mean(MiB)     StdDev   Max(OPs)   Min(OPs)  Mean(OPs)     StdDev    Mean(s) Stonewall(s) Stonewall(MiB) Test# #Tasks tPN reps fPP reord reordoff reordrand seed segcnt   blksiz    xsize aggs(MiB)   API RefNum
write         575.31     575.31     575.31       0.00     287.66     287.66     287.66       0.00  208.58290         NA            NA     0      6   1    1   0     1        1         0    0  10000  2097152  2097152  120000.0 CEPHFS      0
```

(Note that in order to compile this, one still needs to patch libcephfs.h to add `#include <sys/time.h>`. This seems to be a problem that's of course external to IOR)

### DIFF
--- a/src/aiori-CEPHFS.c
+++ b/src/aiori-CEPHFS.c
@@ -18,6 +18,7 @@
 # include "config.h"
 #endif
 
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


JulianKunkel (approved):

JulianKunkel:
Awesome, great.

adilger:
There is no real explanation of what "lazy" means, other than it gives better performance. Could you please provide some basic explanation of the "lazy write" semantics (eg. does it affect data integrity, disable fsync, etc). Also, is this an option that users would be able to use for their applications, or only for benchmarking?

pllopis:
> There is no real explanation of what "lazy" means, other than it gives better performance. Could you please provide some basic explanation of the "lazy write" semantics (eg. does it affect data integrity, disable fsync, etc). Also, is this an option that users would be able to use for their applications, or only for benchmarking?

Indeed @adilger, I should perhaps have added more references to this use case.

In a nutshell, Lazy I/O is a mode that relaxes the POSIX consistency model. The ideas behind this are not new, and different file systems provide different mechanisms (and different apis/semantics) to relax the POSIX consistency model.

In practice, certain workloads collectively write to a single shared file, where each process (e.g. MPI rank) writes to a different (and typically non-overlapping) range.
In this case, Lazy I/O enables having more than one writer per file.
This is exactly the case that is represented in the IO500 benchmark's "hard mode". It's considered "hard" because most POSIX-compliant filesystems (such as CephFS, when Lazy I/O is not enabled) would just make that writing sequential.

On top of that, there are extra calls in order to manage data coherency in the event that buffered writes are combined with readers.
The CephFS documentation on Lazy I/O provides more details about this Lazy I/O mode and how it's implemented: https://docs.ceph.com/en/latest/cephfs/lazyio/

PR #405, opened 2022-01-04
Fixing resource leaks in utilities.c

ryancaicse:
Fixing resource leaks in utilities.c


JulianKunkel:
Alright, merged it, would have been OK to just move the fclose() up some lines.

PR #413, opened 2022-02-08
Enable fsync Support for Implemented Backends

chriswasser:
Hello once more,

While testing features on the main branch, I noticed that the `fsync` option is not correctly enabled for some backends although they implement the functionality (most recently HDF5 and NCMPI as well).This PR fixes only this behavior.

Checking for these `fsync` capabilities at a global level might be useful to keep the amount of warning messages to a minimum but necessitates changes within `ior.c` for development of a new backend. Both variants of handling the `fsync` option have its merits as a backend which does not implement `fsync` could otherwise execute without notifying the user. I just wanted to raise awareness about this.

Similar issues occur for `fsyncPerWrite`. Although `ior.c` explicitly calls `fsync` after each `xfer` if `fsyncPerWrite` is set, some backends also check for `fsyncPerWrite` in their hints causing double flushes. Currently, both `fsync` and `fsyncPerWrite` appear as POSIX-only options in the documentation. After deciding how to permanently handle these options, this should be adjusted accordingly.

Side notes regarding handling of file access flags by I/O backends:
- The HDF5 and NCMPI seem to interpret the file access flags passed by IOR a little differently. While HDF5 is happy to translate `IOR_WRONLY` and `IOR_RDWR` into the same `H5F_ACC_RDWR`, NCMPI warns for `IOR_WRONLY` as there is only `NC_WRITE` within PnetCDF, which also allows reading the file. This leads to warnings being printed for NCMPI for successful executions. Therefore, it might be sensible to set `NC_WRITE` when `IOR_WRONLY` is requested.
- Interestingly, some backends check for `IOR_DIRECT` to show warnings since they do not support it; others don't bother checking for this flag. The backends which allow direct I/O use the hint `direct_io` instead, and thus, never rely on `IOR_DIRECT`, which is never actually set anyhow. This status seems like the product of missing ported code when transitioning between AIORIs.

I hope the above information can spark some discussions or provide starting points for implementing a more uniform behavior of backends within IOR.

Best,

Christian


chriswasser:
If an issue is more suited as a permanent discussion forum, I can move the non-PR related text into its own issue.

PR #457, opened 2023-03-13
userDoc: fix mentioning of IOR STOP and be more specific about case

drhpc:
I noticed IOR END being mentioned in the user documentation about scripting. There is IOR STOP instead.
Furthermore, those control words are parsed case-insensitively, which perhaps should be noted.
Which I did.
Please merge at your leisure.


JulianKunkel:
Awesome, thx, improvements in the doc is always great

PR #458, opened 2023-03-13
lowercase ./ior in tutorial

drhpc:
At least on my system with a case-sensitive file system, ./IOR won't do any good.


glennklockwood:
Thanks! This is back from the early 2.x days when IOR could be compiled in either uppercase or lowercase mode.

PR #459, opened 2023-03-24
ior: improve error for existing output file

adilger:
If the '-E' option is not used and existing output files are found, the current error message says that the files will be "overwritten" but this is not totally accurate.  Without '-E' the file will be deleted first, which makes a difference if the files were precreated with a specific layout.

Update the error to report that files will be deleted in this case.


adilger (commented):
@JulianKunkel @glennklockwood 

What is the process to get this trivial change landed? Twice recently we've had testers burn a day or two of time thinking that "overwrite" meant that their pre-created files (with specific Lustre layout) were being used for testing, only to found IOR silently deleting them before testing started.

PR #473, opened 2024-01-11
Pull DAOS fixes into 4.0 release stream

glennklockwood:
@mchaarawi or @roblatham00, could you test this using DAOS to ensure that these DAOS patches got pulled into the 4.0 release branch correctly?  I want to cut a stable 4.0 release, but I also want it to works with DAOS as it exists today.


mchaarawi (approved):
looks good

PR #480, opened 2024-03-20
WIP: use new MPI big count routines

roblatham00:
Ashwin Reghunandanan from HPE made these changes to IOR to exercise the MPI-4 "large count" routines.  

Hm.  this probably needs an AC_CHECK_TYPE for older MPI implementations without the Count and _c routines, and maybe a bit of pre-processor help.


roblatham00:
updated for a bit of configure checking and backwards compat for older implementations

PR #487, opened 2024-05-07
Adding the option for selecting a lustre pool to the source code.

adrianjhpc:
Adding the option for selecting a lustre pool to the source code. Really only done for the POSIX backend so far but can be expanded beyond.

Let me know if anything specific (tests, documentation, etc...) is needed to help this merge back into the source.


mchaarawi:
this change broke builds where lustre is not installed. here is my attempt to fix it:
https://github.com/hpc/ior/pull/488

adrianjhpc:
Thanks for the fix @mchaarawi

PR #489, opened 2024-05-15
aiori/mpiio: optimize get size by issuing only on 1 rank

mchaarawi:
In the single shared file case, there is no need for all ranks to query the file size; only a single rank can do that and bcast the result to others.

### DIFF
--- a/src/aiori-MPIIO.c
+++ b/src/aiori-MPIIO.c
@@ -650,25 +650,23 @@ IOR_offset_t MPIIO_GetFileSize(aiori_mod_opt_t * module_options, char *testFileN
           return 0;
         IOR_offset_t aggFileSizeFromStat, tmpMin, tmpMax, tmpSum;
         MPI_File fd;
-        MPI_Comm comm;
         MPI_Info mpiHints = MPI_INFO_NULL;
 
-        if (hints->filePerProc == TRUE) {
-                comm = MPI_COMM_SELF;
-        } else {
-                comm = testComm;
+        if (hints->filePerProc || rank == 0) {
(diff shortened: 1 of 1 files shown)


mchaarawi:
any feedback on this one @roblatham00 @JulianKunkel ?

roblatham00:
I'm a big fan of these kinds of optimzations.  You can see this pattern in several places in ROMIO for example (https://github.com/pmodels/mpich/blob/main/src/mpi/romio/adio/common/ad_resize.c)  

"get file size" is like a read... are there any oddball file systems (nfs) where the file size is lazily updated?

src/aiori-MPIIO.c
```diff
@@ -650,25 +650,23 @@ IOR_offset_t MPIIO_GetFileSize(aiori_mod_opt_t * module_options, char *testFileN
...
         IOR_offset_t aggFileSizeFromStat, tmpMin, tmpMax, tmpSum;
         MPI_File fd;
-        MPI_Comm comm;
         MPI_Info mpiHints = MPI_INFO_NULL;
 
-        if (hints->filePerProc == TRUE) {
-                comm = MPI_COMM_SELF;
-        } else {
-                comm = testComm;
+        if (hints->filePerProc || rank == 0) {
```
roblatham00, line 655: I initially thought we needed a sync/barrier/sync here but GetFileSize is called from CheckFileSize , where we just closed the file and issued a barrier.  that seems pretty safe for even the strangest of strange file systems.

JulianKunkel (approved):

Issue #491, opened 2024-05-29
Lustre API error introduced in #488?


johnbent:
Hello,

I'm doing some simple testing of Lustre 2.15.4 with this command:

./src/ior -o /mnt/lustre/testfile -t 1m -b 16m -s 16

And I'm suddenly seeing this error as of #488 :

WARNING: Lustre pool specified, but the Lustre User API on this system does not support that.
ERROR: Setting Lustre pool is not supported for the version of the Lustre API used on this system, (aiori-POSIX.c:219)

When I revert to 54fd197c56dbb7a7706221acbe80198903405456, then the error goes away and this IOR command runs successfully as expected.

@mchaarawi , any ideas? 

And thanks for all the hard work of all the maintainers for this great tool!

mchaarawi:
This check and error was something added in https://github.com/hpc/ior/commit/95334b9f3b68c500672be4317c5655b211c0064a
which is between those 2 versions you had mentioned. 

@adrianjhpc  any idea on this issue? Unfortunately I do not have access to a lustre FS to check this myself.

adrianjhpc:
Definitely my fault. Let me take a look.

johnbent:
Sorry for the false acc @mchaarawi ! Thanks for getting @adrianjhpc involved.

Thanks @adrianjhpc for taking a look!

adrianjhpc:
fyi, I've not forgotten this, just trying to work out the best way to fix it.

adrianjhpc:
@johnbent can you try pull request 492, i.e. `gh pr checkout 492` and see if it works for you?

johnbent:
Thank you very much @adrianjhpc . Unfortunately, I'm very sad to report that recently my entire workflow broke due to the unexpected (/s haha) EOL of CentOS Stream 8 so I'm not currently able to reproduce. I'll definitely will get back to this as soon as I can. Sorry about that; especially after you were so awesomely responsive. I'm assigning this ticket to me at least. :)

adrianjhpc:
No worries, if you do get back to it let me know how it looks.

PR #492, opened 2024-06-08
Fixing issue raised in issue 491 by previously added Lustre pool functionality

adrianjhpc:
To fix https://github.com/hpc/ior/issues/491 I've updated the Lustre pool functionality to make it more robust, particularly on systems where the Lustre API is not being used or when people are building without specific lustre functionality.


adrianjhpc:
@JulianKunkel any chance you can merge this fix in?

adrianjhpc:
Thanks

PR #493, opened 2024-06-20
Fix a couple of minor issues with IOR output

hjelmn:
This commit fixes two issues:

 - The AIO backend prints the max outstanding transactions when starting. This is printed by every MPI rank and doesn't add much. Probably is left over from some debugging.

 - Init MPI ***before*** parsing the command line. Without this every rank thinks it is rank 0 so messages that should only be printed by rank 0 are printed n times.

### DIFF
--- a/src/aiori-aio.c
+++ b/src/aiori-aio.c
@@ -84,7 +84,6 @@ static void aio_initialize(aiori_mod_opt_t * param){
   if(io_setup(o->max_pending, & o->ioctx) != 0){
     ERRF("Couldn't initialize io context %s", strerror(errno));
   }
-  printf("%d\n", (o->max_pending));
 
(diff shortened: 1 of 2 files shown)


hjelmn:
Been a long time since I touched IOR but these two issues have been bugging me. Julian, PTAL.

hjelmn:
The only issue is the whole parse for -h before starting MPI thing. I can probably work around that if we want to maintain that.

JulianKunkel:
The issue with -h is that there exist some MPI implementations that require a *parallel job* to be started before they can be initialized. Thus, it would not be possible to see the arguments of ior without running it in a parallel job.
This in turn might not be possible in an interactive way. 
It is very annoying, therefore, I would not want to merge the "-h" behavior change.
An alternative might be that we have a "ior-help" tool or such without MPI, but it feels awkward. Any suggestions are welcome though.

Issue #506, opened 2025-01-23
mdtest size > first == last hang


jschwartz-cray:
The `-f`/`-l` options are used to restrict the number of tasks mdtest runs to smaller subsets of the total size specified via the job MPI parameters.

In this mode a subset of the ranks will not participate in the test, and those ranks have to be managed properly so they join up with the ranks that did at the end.

The recently refactored logic fixed one issue but created another in the corner case of size > first == last. In this scenario only one rank participates in the test, but all ranks are duping MPI_COMM_WORLD and the barrier behavior is not correct for this scenario resulting in a hang.

The relevant code is here:

```
        if(i < last){
          MPI_Group testgroup;
          range.last = i - 1;
          MPI_Group_range_incl(worldgroup, 1, (void *)&range, &testgroup);
          MPI_Comm_create(world_com, testgroup, &testComm);
          MPI_Group_free(&testgroup);
          if(testComm == MPI_COMM_NULL){
            continue;
          }
        }else{
          MPI_Comm_dup(world_com, & testComm);
        }
```

One solution to this involves making the logic common and ensuring that any ranks which aren't participating are handled in the same manner as they are in https://github.com/hpc/ior/blob/9f97b1069459317304753466e5a9bcf9906b623b/src/ior.c#L117. I will be submitting a PR which implements this fix and some minor error handling improvements as a separate commit.

jschwartz-cray:
https://github.com/hpc/ior/pull/507

JulianKunkel:
Thanks for the checkup and fixes.

PR #507, opened 2025-01-24
fix(#506): mdtest size > first == last hang

jschwartz-cray:
The -f/-l options are used to restrict the number of tasks mdtest runs to smaller subsets of the total size specified via the job MPI parameters.

In this mode a subset of the ranks will not participate in the test, and those ranks have to be managed properly so they join up with the ranks that did at the end.

The recently refactored logic fixed one issue but created another in the corner case of size > first == last. In this scenario only one rank participates in the test, but all ranks were duping MPI_COMM_WORLD and the barrier behavior was not correct for this scenario resulting in a hang.

This solves the problem by making the logic common (a new group and communicator will always be created for the test whether it is for all ranks or a subset) and ensuring that any ranks which aren't participating are handled in the same manner as in ior.c:117.

### DIFF
--- a/src/mdtest.c
+++ b/src/mdtest.c
@@ -286,7 +286,7 @@ static void phase_prepare(){
     system(o.prologue);
   }
   if (o.barriers) {
-    MPI_Barrier(testComm);
+    MPI_CHECK(MPI_Barrier(testComm), "MPI_Barrier error");
   }
 }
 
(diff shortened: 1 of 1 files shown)


JulianKunkel:
I can confirm the hang. 
Thanks for adding the MPI checks.

PR #510, opened 2025-02-26
Fix wrong usage of putenv()

sydidelot:
As explained in the putenv(3) man page, putenv() places the string pointer passed in directly into the environment - no copy.

Given that the hint array is allocated on the stack, this will definitely corrupt the environment. Should use setenv(3) instead, which copies strings.


sydidelot:
Thanks for the merge!

PR #511, opened 2025-02-26
new IOR-Backend that uses libnfs

ggoers:
This Pull Request adds a new implementation of an IOR backend that communicates over NFS using the library "libnfs"(https://github.com/sahlberg/libnfs).

I implemented the new backend as part of my bachelor thesis. The motivation behind this is to provide a way to be more tolerant to crashes of the NFS server during the benchmark. Because the NFS client now runs in user space instead of kernel space, it is in many cases more tolerant if the NFS server crashes. This means that the client node does not have to be rebooted if it stops responding, which makes load tests against the NFS-Server easier.

I have also developed integration tests for my backend implementation, which are located under testing/libnfs. As with all other changes, you can decide if these are welcome here.


JulianKunkel (approved):
Great contribution.

Issue #500, opened 2024-12-02
Question : Task-reorder option -C and -Z in IOR on single server - performance difference.


PSC-cloud:
Hi,
 I am running IOR tests with two different options -C 1 and -Z using 48 processes with one file per process. First the entire data is written (more than a terabyte per file) and then read. Read performance is compared.
 
IOR is being run on ZFS filesystem on a single server. 6 Samsung SSD (EVO990 Pro) are used in RAIDZ0 configuration. 

We see considerably more throughput in -Z option compared to -C 1 option.  Could you please provide some insight whey random ordering of the tasks in IOR (-Z) compared to constant ordering of the tasks (-C 1) achieves more throughput in all cases provided here.  

<img width="515" alt="image" src="https://github.com/user-attachments/assets/29d72526-8112-4c64-856f-5ba7fe51549e">

I am trying to figure out on a single server how random reorder of tasks compared to constant reorder in IOR makes the difference.

carns:
Just a guess, but part of the reason may be the randomization method.  Looking at https://github.com/hpc/ior/blob/51d8a790153f13e4637e2483f5947b1aedb78e4d/src/ior.c#L1366 , it looks like the offset is independently randomized at each rank, meaning that some ranks may read the same data.  It is not a "shuffle" in which all offsets are covered exactly but just in a different permutation, as in the -C option.

If you are using Darshan you can `export DXT_ENABLE_IO_TRACE=1` to capture an exact trace of the I/O operations and see how the two modes compare in terms of what offsets are being read by which ranks.

JulianKunkel:
That is correct, the rankOffset is randomized until it exceeds at least the number of tasks per Node to prevent the read caching. The process then pretends to be the new rank. This leads to cases where multiple ranks pretend to be the same rank.  If you use Verbosity Level > 2, it prints stats how many ranks hit which rankOffset.
I suppose with a single server the read cache of ZFS helps to increase performance in your case. 
Have you been using one client node?

adilger:
Julian, one of the suggestions at the SC'24 IO500 BOF was that all ranks open all files during the random read phase, and then read random blocks from all files randomly.  This gives the largest input dataset, and will also work consistently if the ior-easy-write phase does not use file-per-process mode.

JulianKunkel:
For a singled shared file, it is useful to randomize the reads. 
It is not useful to open all files on all procs. Even on 1000 ranks, this quickly becomes the performance issue to open/close the files.

[...]

PR #512, opened 2025-03-11
Update behaviour of reorderTasksRandom to shuffle ranks and not just rankOffsets.

celrick:
With the current behaviour, when specifying `reorderTasksRandom` multiple times with `-Z -Z`, the offsets for readback are shuffled, but since the rank that does the read operation is decided by `rank + rankOffset` the result is that multiple ranks can/will end up reading the same data. This can lead to a (superficial) increase to the reported read performance as data can be cached on a client and given to multiple ranks.

Instead, if we shuffle the `rankOffset` as is done currently and then subtract the current rank, this will ensure that the ranks that do the read operations are a true permutation of the write ranks.

This is related to https://github.com/hpc/ior/issues/500

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -1365,7 +1365,7 @@ static void TestIoSys(IOR_test_t *test)
                                         file_hits_histogram(params);
                                 }
                         }
-                        if (params->reorderTasksRandom > 1) { /* Shuffling rank offset */
+                        if (params->reorderTasksRandom > 1) { /* Shuffling ranks for read back */
                                 int nodeoffset;
                                 unsigned int iseed0;
                                 nodeoffset = params->taskPerNodeOffset;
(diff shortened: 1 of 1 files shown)


JulianKunkel:
Hi, there are two options, shuffling -> here every process gets another rank number but randomly assigned. This won't produce an overlap.
Second option is to create a random assignment -> this may lead to two processes with the same rank. 
Could you elaborate on the exact issue based on an example, and why this is not fixed using -Z once?

celrick:
So to elaborate on the exact issue:

With both options `-Z` and `-Z -Z` the `rankOffset` variable is either randomly assigned or randomly shuffled as you say.

So e.g. in a job with 3 tasks with the `shuffle option`

Rank 0 could end up with `rankOffset` of 2 and end up performing read on the data written by rank 2
Rank 1 could end up with `rankOffset`  of 1 and end up performing read on the data written by rank 2
Rank 2 could end up with `rankOffset` of 0 and end up performing read on the data written by rank 2.

So even though the `rankOffset`s are properly shuffled (each value appears once and once only), the tasks that ultimately perform the read operation are not shuffled (and in the example above all read the same data)

JulianKunkel:
Thank you, I could reproduce the issue and indeed your fix is the easiest albeit a bit counter intuitive, therefore, I added some descriptions.

PR #519, opened 2025-04-30
WIP: Delete file to trigger any "first touch" calls

roblatham00:
Some storage systems (DAOS, PVFS, possibly others) do a "library init" the very first time a program calls MPI_File_Open or MPI_File_delete.

In some ways it's entirely fair to include these first touch time but it can also obscure the effects of optimizations or tuning to have one of the iterations consistently take 10x longer.

Since I don't know if this is a good idea or not, I thought I'd fire off this pull request to see if the community had any thoughts

### DIFF
--- a/src/aiori-MPIIO.c
+++ b/src/aiori-MPIIO.c
@@ -626,8 +626,7 @@ void MPIIO_Delete(char *testFileName, aiori_mod_opt_t * module_options)
   mpiio_options_t * param = (mpiio_options_t*) module_options;
   if(hints->dryRun)
     return;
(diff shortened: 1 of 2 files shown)


adilger (commented):
I think this should be a configurable option, and IMHO not enabled by default. If the initialization time is so significant that it is affecting the benchmark time, then it is also likely significant enough that it affects application runtime and should be measured and not hidden.  

Maybe the initialization should be printed as part of the test output?

roblatham00:
Yeah, definitely should be an option.  I should be running each iteration/config of the benchmark for tens of seconds anyway, so a one second initialization shouldn't be significant... Just takes a lot longer to parameter sweep (hours, days?) when doing it the right way.

PR #523, opened 2025-05-11
Attempt for 64 Bit rand.

JulianKunkel:
See https://github.com/IO500/io500/issues/88#issuecomment-2869874292

### DIFF
--- a/src/utilities.h
+++ b/src/utilities.h
@@ -82,5 +82,5 @@ void *aligned_buffer_alloc(size_t size, ior_memory_flags type);
 void aligned_buffer_free(void *buf, ior_memory_flags type);
 
 void srand64(uint64_t state);
-uint64_t rand64();
+uint64_t rand64(void);
 #endif  /* !_UTILITIES_H */
--- a/src/utilities.h
+++ b/src/utilities.h
@@ -80,4 +80,7 @@ char * PrintTimestamp(void); // TODO remove this function
 unsigned long GetProcessorAndCore(int *chip, int *core);
 void *aligned_buffer_alloc(size_t size, ior_memory_flags type);
 void aligned_buffer_free(void *buf, ior_memory_flags type);
+
+void srand64(uint64_t state);
+uint64_t rand64();
 #endif  /* !_UTILITIES_H */
--- a/src/aiori-MPIIO.c
+++ b/src/aiori-MPIIO.c
@@ -506,9 +506,9 @@ static IOR_offset_t MPIIO_Xfer(int access, aiori_fd_t * fdp, IOR_size_t * buffer
                                  * deal with us reporting that we wrote N times more
                                  * data than requested. */
                                 length = hints->transferSize;
-                                MPI_CHECK(MPI_Get_elements_x(&status, MPI_INT, &elementsAccessed),
+                                MPI_CHECK(MPI_Get_elements_x(&status, MPI_BYTE, &elementsAccessed),
                                           "can't get elements accessed" );
-                                xferBytes = elementsAccessed * sizeof(MPI_BYTE);
+                                xferBytes = elementsAccessed;
                         }
                 } else {
                         /*
@@ -548,11 +548,11 @@ static IOR_offset_t MPIIO_Xfer(int access, aiori_fd_t * fdp, IOR_size_t * buffer
                                                   "cannot access explicit, noncollective");
                                 }
                         }
-                        MPI_CHECK(MPI_Get_elements_x(&status, MPI_INT, &elementsAccessed),
+                        MPI_CHECK(MPI_Get_elements_x(&status, MPI_BYTE, &elementsAccessed),
                                    "can't get elements accessed" );
 
                         expectedBytes = length;
-                        xferBytes = elementsAccessed * sizeof(MPI_BYTE);
+                        xferBytes = elementsAccessed;
                 }
 
                 /* Retrying collective xfers would require syncing after every IO to check if any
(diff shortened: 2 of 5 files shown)


src/utilities.h
```diff
@@ -80,4 +80,7 @@ char * PrintTimestamp(void); // TODO remove this function
 unsigned long GetProcessorAndCore(int *chip, int *core);
 void *aligned_buffer_alloc(size_t size, ior_memory_flags type);
 void aligned_buffer_free(void *buf, ior_memory_flags type);
+
+void srand64(uint64_t state);
+uint64_t rand64();
```
Obihoernchen, line 85: Should be:

```c
uint64_t rand64(void);
```

Otherwise the following error is thrown:

```
./build/ior/src/utilities.h:85:1: error: function declaration isn't a prototype [-Werror=strict-prototypes]
   85 | uint64_t rand64();
      | ^~~~~~~~
```

Obihoernchen (changes requested):

src/utilities.h
Obihoernchen: It doesn't work for me (reads 0 bytes).

JulianKunkel:
Try again, please.

Obihoernchen:
Works now but doesn't change much:

Quick test with 10 nodes, 12 npernode (1200 total procs), 60s stonewallingTime.
Files from ior-easy-write are 36.62 TiB aggregated.

```
Options: 
api                 : POSIX
apiVersion          : 
test filename       : /lustre/io500/2025.05.14-21.40.16/ior-easy/ior_file_easy
access              : file-per-process
type                : independent
segments            : 1000000
ordering in a file  : random
ordering inter file : constant task offset
task offset         : 1
nodes               : 10
tasks               : 1200
clients per node    : 120
memoryBuffer        : CPU
dataAccess          : CPU
GPUDirect           : 0
repetitions         : 1
xfersize            : 4096 bytes
blocksize           : 4096 bytes
aggregate filesize  : 4.47 TiB
stonewallingTime    : 60
stoneWallingWearOut : 0

Results: 

access    bw(MiB/s)  IOPS       Latency(s)  block(KiB) xfer(KiB)  open(s)    wr/rd(s)   close(s)   total(s)   iter
------    ---------  ----       ----------  ---------- ---------  --------   --------   --------   --------   ----
WARNING: Expected aggregate file size       = 33554432000
WARNING: Stat() of aggregate file size      = 40265318400000
WARNING: Using actual aggregate bytes moved = 4915200000000
WARNING: Maybe caused by deadlineForStonewalling
read      987239     253219159  0.000004    4.00       4.00       0.006183   4.74       0.002975   4.75       0   
```

Very same test with `--posix.odirect` (limited by client performance).

```
access    bw(MiB/s)  IOPS       Latency(s)  block(KiB) xfer(KiB)  open(s)    wr/rd(s)   close(s)   total(s)   iter
------    ---------  ----       ----------  ---------- ---------  --------   --------   --------   --------   ----
WARNING: Expected aggregate file size       = 33554432000
WARNING: Stat() of aggregate file size      = 40265318400000
WARNING: Using actual aggregate bytes moved = 1076395753472
WARNING: Maybe caused by deadlineForStonewalling
read      17105      4379520    0.000284    4.00       4.00       0.007158   60.00      0.003510   60.01      0   
```

JulianKunkel:
I realize, that since we use 1 file in easy per proc, the rand-32 wasn't in fact a limit. The IOR rand64 is still an improvement. I also conducted statistical test for randomness (in R), which the new random function holds.
This line indicated the actual file size found for rank 0 process (the label is misleading as it wasn't aggregated):
Expected file size       = 33554432000
Given that each process determines the file size by itself (we hadn't used wear-out but the benchmark finishes within 5s so it was like wear-out), all ranks should have that file size, this is confirmed by looking at 1200 procs having:
Stat() of aggregate file size      = 40265318400000

Thus each files has a size of 31 GiB. 120 ranks run per node  = 3720 GiB data to pick from.
Given RAM of the machines was about 700 GiB, Cache hit rate of ~18% is expected, where this is obviously more.

After another code review, I think that the way the existing file size was calculated was wrong, thus only 300 MiB of data was read. Please try again.

Obihoernchen:
051dc84 crashes for me:

```
IO500 version io500-isc25_v1 (standard)
ERROR INVALID (src/phase_dbg.c:15) stonewall-time < 300s
io500: src/phase_ior.c:9: ior_process_write: Assertion `res' failed.
[h21-04:94653] *** Process received signal ***
[h21-04:94653] Signal: Aborted (6)
[h21-04:94653] Signal code:  (-6)
io500: src/phase_ior.c:9: ior_process_write: Assertion `res' failed.
[h21-02:92426] *** Process received signal ***
[h21-02:92426] Signal: Aborted (6)
[h21-02:92426] Signal code:  (-6)
io500: src/phase_ior.c:9: ior_process_write: Assertion `res' failed.
[h23-01:95815] *** Process received signal ***
[h23-01:95815] Signal: Aborted (6)
[h23-01:95815] Signal code:  (-6)
io500: src/phase_ior.c:9: ior_process_write: Assertion `res' failed.
[h23-02:95210] *** Process received signal ***
[h23-02:95210] Signal: Aborted (6)
[h23-02:95210] Signal code:  (-6)
```

Obihoernchen:
Works now. See: https://github.com/IO500/io500/pull/91#issuecomment-2885114308

Obihoernchen (approved):

PR #535, opened 2026-03-05
feat #534: Add task to processor mapping to output

jschwartz-cray:


src/ior-output.c
```diff
@@ -409,6 +409,54 @@ void ShowTestEnd(IOR_test_t *tptr){
...
+        char processor[MAX_STR];
+        int processorLen = MAX_STR;
+        char task[MAX_STR];
+
+        MPI_CHECK(MPI_Get_processor_name(processor, &processorLen),
+                                         "cannot get processor name");
+
+        if (rank > 0) {
+                /* wait for the previous rank to signal us */
+                MPI_CHECK(MPI_Recv(NULL, 0, MPI_BYTE, rank - 1, 0,
```
JulianKunkel, line 426: Actually, you do not need this. You can simply send the data to rank 0. 
Try :-)

src/ior-output.c
```diff
@@ -409,6 +409,54 @@ void ShowTestEnd(IOR_test_t *tptr){
...
+                                   params->testComm, MPI_STATUS_IGNORE), "recv error");
+
+                /* send our processor assignment to rank 0 */
+                MPI_CHECK(MPI_Send(processor, processorLen, MPI_CHAR, 0, 1,
+                                   params->testComm), "send error");
+        }
+
+        if (rank < params->numTasks - 1) {
+                /* signal the next rank */
+                MPI_CHECK(MPI_Send(NULL, 0, MPI_BYTE, rank + 1, 0,
```
JulianKunkel, line 436: Actually, you do not need this. You can simply send the data to rank 0. 
Try :-)

JulianKunkel (commented):
Please check out these changes - no MPI_Send() Recv to previous rank needed, then I will approve
Good job!

src/ior-output.c
jschwartz-cray: > Actually, you do not need this. You can simply send the data to rank 0. Try :-)

Yeah, the reason for doing it this way is to get the tasks to come out in numeric order as otherwise it's just a race and they come out in a jumbled order.

The old version of this code used a more complicated bubble sort to accomplish something similar, though that also didn't work very well since they were ordered lexicographically instead of strictly numerically (so rank 2 and 20 showed up together, etc.)

I figured since this is a one-time thing at startup it doesn't hurt anything to synchronize the tasks like this to get the task output in-order.

PR #539, opened 2026-07-06
refactor: extract cuFile/GPU runtime into abstraction layers

potatogim:
cuFile calls were inlined directly in aiori-POSIX.c. Split them out:

- gpu_runtime.h / cuda_runtime.c: cudaMalloc, cudaFree, cudaMemcpy, etc.
- gpu_io.h / gpu_io_cufile.c: cuFileDriverOpen, cuFileHandleRegister,
  cuFileRead/Write with error decoding

configure.ac:
- --with-cuda=PREFIX with save/restore (was leaking into global flags)
- --with-gpu-direct-backend=cufile|none replaces --with-gpuDirect
  (old flag still works as alias)
- --with-cufile=PREFIX

No behaviour change.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -78,69 +78,135 @@ AS_IF([test "$ac_cv_header_gpfs_h" = "yes" -o "$ac_cv_header_gpfs_fcntl_h" = "ye
     ])
 ])
 
-# Check for CUDA
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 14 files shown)


JulianKunkel (approved):
Thanks for the code contribution. It looks reasonable. I do trust you tested on different platforms.

PR #543, opened 2026-09-05
Add --memoryAlign to allocate huge-page-backed I/O buffers

paf0186:
### Problem

`aligned_buffer_alloc()` aligns I/O buffers to the system page size, which is all `O_DIRECT` requires. Whether the buffer then ends up huge-page-backed is left entirely to the system THP policy and to where the allocator happened to land.

Measured on a 5.14 kernel, `AnonHugePages` over the buffer's VMA:

| buffer | `transparent_hugepage=always` | `transparent_hugepage=madvise` |
|---|---|---|
| 4 / 16 / 64 MiB, stock 4KiB alignment | **already 100%** | 0% |
| 1 MiB, stock 4KiB alignment | 0% | 0% |
| 2MiB-aligned + `MADV_HUGEPAGE` | 100% | **100%** |

So on the common `always` policy a large buffer is already fully backed, and this option changes nothing for it. The two cases it addresses are the other rows:

- Under the `madvise` policy nothing is backed, at any size.
- Below the PMD size nothing is backed under either policy: a sub-PMD buffer needs a multi-size THP, and those require the region to be *naturally aligned* to the folio order, which a 4KiB-aligned buffer is not.

That matters when benchmarking a data path that behaves differently for huge-page-backed user buffers -- one that pins user pages and coalesces runs of pages belonging to the same folio, say, or one registering buffers for RDMA. Today IOR cannot *ask* for such a buffer. It gets one by luck of a global system setting the benchmark does not record, and cannot get one at all for small transfers or on a `madvise`-policy machine.

I hit this measuring a Lustre change that collapses per-page bookkeeping when the user buffer is folio-backed.

### Change

`--memoryAlign=SIZE` aligns I/O buffers to the given boundary and, on Linux, advises `MADV_HUGEPAGE` over the buffer. Aligning to 2MiB is naturally aligned for every smaller order too, so one option covers PMD THP and every mTHP order.

**The default is unchanged**: 0 means the system page size and no `madvise()` call.

Two details that turned out to matter:

- The `madvise()` has to happen **before** the buffer is first touched. Once a page has been faulted in as a base page, only khugepaged can collapse it, which is asynchronous and may not happen during a short run. `safeMalloc()` zeroes the buffer immediately, so this path now uses `malloc()` plus an explicit `memset()` after the `madvise()`, keeping the same zeroing contract.
- Only the aligned region is advised, since `madvise()` requires a page-aligned start and the `malloc()`ed pointer is not one.

Rejected at validation: non-power-of-two, below the page size, above 1g.

### Testing

`thp_fault_alloc` delta across an IOR run with `transparent_hugepage=madvise`, so `MADV_HUGEPAGE` is required rather than merely helpful:

| transfer | default | `--memoryAlign=2m` |
|---|---|---|
| 4m | 0 | 2 |
| 16m | 0 | 8 |

Counts match the transfer size exactly (4MiB = 2 huge pages, 16MiB = 8), so the whole buffer is backed rather than part of it.

`strace` confirms the advised length is correct at sub-huge-page transfer sizes, where rounding the length down to a huge-page multiple would pass 0:

```
t=64k  --memoryAlign=2m  madvise(0x1c00000, 65536, MADV_HUGEPAGE) = 0
t=1m   --memoryAlign=2m  madvise(0x1600000, 1048576, MADV_HUGEPAGE) = 0
t=4m   --memoryAlign=2m  madvise(0x1200000, 4194304, MADV_HUGEPAGE) = 0
t=*    default           <no madvise call>
```

`O_DIRECT` works in both modes.

On the VM used here, write bandwidth over 5 iterations is unchanged: medians 2418 vs 2341 MiB/s, well inside that setup's run-to-run spread. **That is a no-regression check, not a measure of the benefit.** On real hardware, write bandwidth is much faster with huge-page-backed buffers in some cases -- which is the point of the option. A VM whose storage is the bottleneck cannot show that.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ior.c
+++ b/src/ior.c
@@ -906,6 +906,7 @@ static void InitTests(IOR_test_t *tests)
 static void XferBuffersSetup(IOR_io_buffers* ioBuffers, IOR_param_t* test,
                              int pretendRank)
 {
+        aligned_buffer_set_alignment((size_t) test->memoryAlign);
         ioBuffers->buffer = aligned_buffer_alloc(test->transferSize, test->gpuMemoryFlags);
 }
 
@@ -1478,6 +1479,15 @@ static void ValidateTests(IOR_param_t * test, MPI_Comm com)
         IOR_param_t defaults;
         init_IOR_Param_t(&defaults, com);
 
+        if (test->memoryAlign != 0) {
+          uint64_t pageSize = (uint64_t) sysconf(_SC_PAGESIZE);
+          if (test->memoryAlign & (test->memoryAlign - 1))
+            ERR("memoryAlign must be a power of two");
+          if (test->memoryAlign < pageSize)
+            ERR("memoryAlign must be at least the system page size");
+          if (test->memoryAlign > (uint64_t) 1 << 30)
+            ERR("memoryAlign is unreasonably large (> 1g)");
+        }
         if (test->gpuDirect && test->gpuMemoryFlags == IOR_MEMORY_TYPE_CPU )
           ERR("GPUDirect requires a non-CPU memory type");
         if (test->gpuMemoryFlags == IOR_MEMORY_TYPE_GPU_DEVICE_ONLY && ! test->gpuDirect )
--- a/src/ior.h
+++ b/src/ior.h
@@ -135,6 +135,7 @@ typedef struct
     int randomSeed;                  /* random seed for write/read check */
     unsigned int incompressibleSeed; /* random seed for incompressible file creation */
     int randomOffset;                /* access is to random offsets */
+    uint64_t memoryAlign;            /* alignment of I/O buffers, 0 = page size */
     size_t memoryPerTask;            /* additional memory used per task */
     size_t memoryPerNode;            /* additional memory used per node */
     char * memoryPerNodeStr;         /* for parsing */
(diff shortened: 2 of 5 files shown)


Obihoernchen:
I found this by accident and did some benchmarks on real HW just as a reference point.

Alma 9 with Kernel 5.14 as well and transparent_hugepage=madvise.
3 interleaved repetitions per configuration, mean with min to max in brackets, GiB/s.

| transfer | write, default | write, `--memoryAlign=2m` | read, default | read, `--memoryAlign=2m` | read gain |
|---|---|---|---|---|---|
| 1m | 47.1 (46.1 to 48.4) | 47.3 (47.0 to 47.4) | 178.7 (177.2 to 181.7) | 178.7 (171.2 to 182.9) | none |
| 4m | 78.5 (78.1 to 79.0) | 78.4 (78.0 to 78.9) | 163.9 (152.0 to 174.2) | 182.7 (182.5 to 183.0) | +11% |
| 16m | 112.7 (112.3 to 113.4) | 112.2 (112.0 to 112.3) | 159.8 (155.3 to 163.1) | 174.7 (174.3 to 175.0) | +9% |
| 64m | 119.5 (118.5 to 121.3) | 118.1 (116.9 to 118.8) | 113.0 (111.9 to 114.4) | 133.8 (130.5 to 135.5) | +18% |

Separate run at 16m, 8 GiB per rank, comparing against the system-wide policy:

| configuration | write | read |
|---|---|---|
| `madvise`, default alignment | 111.6 | 161.3 |
| `madvise`, `--memoryAlign=2m` | 111.5 | 173.5 |
| `always`, default alignment | 111.1 | 173.5 |

**Observations:**
- O_DIRECT reads gain 9 to 18% once the buffer is huge-page-backed, and run-to-run spread shrinks a lot (4m: 22 GiB/s range down to 0.5). Writes are unchanged here, but they are server-bound on this system.
- `--memoryAlign=2m` under `madvise` matches `always` exactly, so the option does what it says: the benefit no longer depends on the system THP policy.
- Backing confirmed via `AnonHugePages`: the difference between arms is exactly 384 x transfer size.
- At 1m there is no effect because this 5.14 kernel has no multi-size THP (no `/sys/kernel/mm/transparent_hugepage/hugepages-*`). A 1 MiB buffer cannot be huge-page-backed there even with the option.

PR #545, opened 2026-09-09
Start of LFSR porting from IO500 testing.

JulianKunkel:
New random permutation method - porting from Jay's IO500 code.

### DIFF
--- a/src/test/rand.c
+++ b/src/test/rand.c
@@ -1,0 +1,15 @@
+#include <stdio.h>
+
+#include <utilities-rand.h>
+
+int main(){
+
+  LFSRRange * range = u_lfsr_range_init (1024+512, 1); // with 1024
+
+  for (int i=0; i < 100; i++){
+    uint64_t pos = u_lfsr_range_step (range);
+    printf("%lld\n", (long long unsigned) pos);
+  }
+  
+  return 0;
+}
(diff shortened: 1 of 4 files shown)


src/test/rand.c
```diff
@@ -0,0 +1,15 @@
+#include <stdio.h>
+
+#include <utilities-rand.h>
+
+int main(){
+
+  LFSRRange * range = u_lfsr_range_init (1024+512, 1); // with 1024
```
JulianKunkel: if I go with 1023, code fails by running into "dummies".

src/utilities-rand.c
```diff
@@ -0,0 +1,206 @@
...
+            printf ("[0, 1, 2, 3]: %d, %d, %d, %d", lfsr->taps[0], lfsr->taps[1], lfsr->taps[2], lfsr->taps[3]);
+            break;
+    }
+    printf (" file_base_offset: %llu\n", (long long unsigned) file_base_offset);
+}
+
+LFSRRange * u_lfsr_range_init (uint64_t blocks, unsigned seed){
+  LFSRRange * range = malloc(sizeof(LFSRRange)); // todo safe_malloc()
+  memset(range, 0, sizeof (LFSRRange));
+  uint64_t b = blocks;
```
JulianKunkel: if I do uint64_t b = blocks & (~(uint64_t) 7);
then I can avoid to use dummies and make the code work.

